PR #419, opened 2026-02-07
feat: add runtime context hint to session greeting

marcusquinn:
## Summary

- Adds app-specific runtime context (config path) to the session greeting output so the AI model knows where settings are stored without guessing
- Includes managed template (`templates/opencode-config-agents.md`) for the OpenCode config-level AGENTS.md greeting instructions
- Also updates the greeting cache for agents without Bash access

Tested: new sessions now show `You are running in OpenCode. Global config: ~/.config/opencode/opencode.json` in the greeting context.

## Summary by CodeRabbit

* **Chores**
  * Updated agent initialization with improved runtime context detection
  * Enhanced caching mechanisms for developer tool configuration and session management

### DIFF
--- a/.agents/scripts/aidevops-update-check.sh
+++ b/.agents/scripts/aidevops-update-check.sh
@@ -190,10 +190,28 @@ main() {
     
     echo "$output"
     
+    # Output runtime context hint for the AI model
+    local runtime_hint=""
+    case "$app_name" in
+        OpenCode)
+            runtime_hint="You are running in OpenCode. Global config: ~/.config/opencode/opencode.json"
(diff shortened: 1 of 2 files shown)


marcusquinn:
@coderabbitai resolve

Fixed the UPDATE_AVAILABLE format mismatch in the template (now shows 4 fields matching the script).

The other two items:
- **Early return skipping hint**: By design - the update banner is a distinct flow. The hint is for normal sessions.
- **Claude Code config path**: Moot - we only run in OpenCode. The Claude Code path is a fallback that may never be used.

PR #548, opened 2026-02-08
fix: guard ((var++)) with || true to prevent silent exit under set -e

alex-solovyev:
## Summary

- Fix `((var++))` arithmetic expressions that silently abort scripts under `set -e` when the variable is 0
- In bash, `((0))` returns exit code 1, and `set -e` (errexit) treats this as a fatal error
- This caused `setup.sh --non-interactive` to abort at `cleanup_deprecated_mcps()` before reaching `deploy_aidevops_agents()`, leaving deployed agents at a stale version

## Changes

- **setup.sh**: Fixed 33 instances of `((var++))` → `((var++)) || true`
- **10 helper scripts**: Fixed the same pattern in scripts that use `set -e`:
  - `anti-detect-helper.sh`, `codacy-cli-chunked.sh`, `crawl4ai-examples.sh`
  - `linter-manager.sh`, `localhost-helper.sh`, `markdown-formatter.sh`
  - `quality-cli-manager.sh`, `setup-linters-wizard.sh`, `setup-local-api-keys.sh`
  - `toon-helper.sh`

## Testing

- `bash -n` syntax check passes on all 11 modified files
- `setup.sh --non-interactive` now runs to completion (previously silently aborted)
- Verified agents deploy correctly with VERSION file synced

## Root Cause

```bash
set -euo pipefail
local cleaned=0
# ... inside a loop when first match found:
((cleaned++))  # cleaned is 0, post-increment returns 0, (( 0 )) = exit 1 → script dies
```

Closes t152

## Summary by CodeRabbit

* **Chores**
  * Enhanced error handling robustness across shell scripts to prevent unexpected termination during script execution.

### DIFF
--- a/.agents/scripts/anti-detect-helper.sh
+++ b/.agents/scripts/anti-detect-helper.sh
@@ -1096,7 +1096,7 @@ show_status() {
     # Profiles
     local profile_count=0
     for dir in "$PROFILES_DIR"/{persistent,clean,warmup}/*/; do
-        [[ -d "$dir" ]] && [[ "$(basename "$dir")" != "default" ]] && ((profile_count++)) || true
+        [[ -d "$dir" ]] && [[ "$(basename "$dir")" != "default" ]] && ((profile_count++)) || true || true
     done
     echo -e "  Profiles:          ${GREEN}$profile_count${NC} configured"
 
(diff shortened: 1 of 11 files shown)


alex-solovyev:
Reopening: task was falsely marked complete in TODO.md, triggering auto-close via issue-sync. Reverted in PR #616.

PR #1280, opened 2026-02-12
t1003: Block parent task completion when any subtask is open

marcusquinn:
## Summary

- Remove `#plan` tag requirement from subtask completion guard — now ALL parent tasks are blocked from `[x]` if any subtask is still `[ ]`
- Add the same guard to `task-complete-helper.sh` (interactive path) — previously only the supervisor path had this check
- Clear error messages listing which subtasks are still open

## Root Cause

The original guard (t278) only checked tasks tagged with `#plan`. But any parent task with subtasks should follow the same rule: a parent can't be resolved until all subtasks are resolved. Without the `#plan` tag, parents could be marked complete while subtasks were still open.

## Changes

1. **`supervisor/todo-sync.sh`** (`update_todo_on_complete()`): Removed `&& "$task_line" == *"#plan"*` condition. Now checks all tasks for open subtasks.
2. **`task-complete-helper.sh`** (`complete_task()`): Added the same awk-based subtask scan. Blocks with clear error listing open subtasks.

## Testing

- `bash -n` syntax check: pass on both files
- ShellCheck: clean (only SC1091 info)
- Functional tests:
  - Parent with open subtask: **blocked** (exit 1, lists open subtask)
  - Simple task (no subtasks): **completes** (exit 0)
  - Parent with all subtasks done: **completes** (exit 0)

Closes #1279

### DIFF
--- a/.agents/scripts/supervisor/todo-sync.sh
+++ b/.agents/scripts/supervisor/todo-sync.sh
@@ -372,13 +372,12 @@ update_todo_on_complete() {
 		return 1
 	fi
 
-	# t278: Guard against marking #plan tasks complete when subtasks are still open.
-	# A #plan task is a parent that was decomposed into subtasks. It should only be
-	# marked [x] when ALL its subtasks are [x]. This prevents decomposition workers
-	# from prematurely completing the parent.
+	# t278, t1003: Guard against marking parent tasks complete when subtasks are open.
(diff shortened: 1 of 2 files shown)


marcusquinn:
Closing as duplicate. PR #1281 (merged) covers the same changes plus pre-commit hook updates.

Issue #1793, opened 2026-02-18
t1185: Fix AI Actions Pipeline 'expected array, got empty' parsing error


marcusquinn:
Task created via claim-task-id.sh

marcusquinn:
Task t1185 cancelled (was: reconcile). Reason: duplicate of t1184 (PR #1797)

PR #1802, opened 2026-02-18
t1185: Fix AI Actions Pipeline 'expected array, got empty' parsing error

marcusquinn:
Fix AI Actions Pipeline 'expected array, got empty' parsing error.

## Changes

- **ai-actions.sh**: Add whitespace-only response check (catches responses that are all whitespace but not empty string)
- **ai-actions.sh**: Add ANSI-strip recovery in array type check — if `jq 'type'` fails on the action plan, strip ANSI escape codes and retry before giving up
- **ai-actions.sh**: Enhance debug logging with hex dump of first 32 bytes for future parse failure diagnosis

## Root Cause

The error `'expected array, got '` (empty plan_type) occurs when `jq 'type'` fails on the action plan returned by `run_ai_reasoning`. This happens when:

1. Two concurrent pulse runs bypass the lock file race window
2. Both call the AI CLI (opencode with `--format default`)
3. The response contains ANSI escape codes that corrupt JSON parsing
4. `extract_action_plan` fails all parse attempts and returns empty
5. `run_ai_reasoning` returns rc=1 with error object — but in some cases ANSI-contaminated content slips through

The `ai-reason.sh` already has ANSI stripping (t1182). This PR adds a second layer of defence in `run_ai_actions_pipeline` itself.

## Fixes

1. **Strip markdown code fences**: Already handled by `extract_action_plan` (Try 2-5 in ai-reason.sh, t1182)
2. **Handle empty/whitespace-only responses**: Added whitespace-trimmed check in pipeline + already in ai-reason.sh (t1182)
3. **Log raw AI response on parse failure**: Enhanced with hex dump for binary/ANSI diagnosis

Ref #1793

### DIFF
--- a/.agents/scripts/supervisor/ai-actions.sh
+++ b/.agents/scripts/supervisor/ai-actions.sh
@@ -1594,8 +1594,10 @@ run_ai_actions_pipeline() {
 		return 1
 	fi
 
-	# Handle empty output — concurrency guard or other silent skip
-	if [[ -z "$action_plan" ]]; then
+	# Handle empty or whitespace-only output — treat as no actions (t1185)
+	local action_plan_trimmed
+	action_plan_trimmed=$(printf '%s' "$action_plan" | tr -d '[:space:]')
+	if [[ -z "$action_plan_trimmed" ]]; then
 		log_info "AI Actions Pipeline: reasoning returned empty output (skipped)"
 		echo '{"executed":0,"failed":0,"skipped":0,"actions":[]}'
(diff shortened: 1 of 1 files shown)


marcusquinn:
Closing as duplicate of #1797 (t1184). Both fix the same 'expected array' parsing bug. #1797 is more comprehensive (covers both ai-actions.sh and ai-reason.sh with whitespace handling, type-checking, indented bracket matching, and full raw response logging).

Issue #1951, opened 2026-02-19
t1251: Investigate frequent stale-evaluating recovery pattern in supervisor


marcusquinn:
**Task ID:** `t1251` | **Status:** open | **Estimate:** `~2h`
**Assignee:** @marcusquinn | **Started:** 2026-02-19T09:04:58Z
**Tags:** `bugfix` `auto-dispatch` `self-improvement`

## Description

Investigate frequent stale-evaluating recovery pattern in supervisor

---
*Synced from TODO.md by issue-sync-helper.sh*

marcusquinn:
## Priority Adjustment Recommendation

**Task:** t1251
**Recommended priority:** high
**Reasoning:** 73% of the last 15 completed tasks required 'Stale evaluating recovery (Phase 0.7)' intervention. This is a systemic issue adding ~35min avg latency per affected task. At current throughput (15 tasks/day), that is ~6 hours of wasted time daily. The investigation task exists but should be prioritized to reduce this overhead.

This is a recommendation from the AI supervisor. A human should review and decide whether to act on it.

---
*Recommended by AI Supervisor (automated reasoning cycle)*

marcusquinn:
Verified: PR #1952 (state:MERGED merged:2026-02-19T09:29:46Z files:2). Task t1251: deploying -> deployed

marcusquinn:
Resolved by PR #1952 (t1251: Fix frequent stale-evaluating recovery — heartbeat, configurable timeout, PR fast-path).

PR #1952, opened 2026-02-19
t1251: Fix frequent stale-evaluating recovery — heartbeat, configurable timeout, PR fast-path

marcusquinn:
WIP - incremental commits

Root causes of 73% stale-evaluating rate:
1. Hardcoded 60s AI eval timeout — too short under load
2. No heartbeat during eval — Phase 0.7 could not distinguish active eval from dead eval
3. Phase 1 called AI eval even when PR was already in DB (wasted 60-90s)

Fixes:
- Configurable eval timeout via SUPERVISOR_EVAL_TIMEOUT (default 90s)
- Heartbeat: evaluate_with_ai() updates updated_at at start of each eval
- New root cause: eval_in_progress_heartbeat_Ns — Phase 0.7/1c skip recovery
- Phase 1 fast-path: skip AI eval when PR already in DB (t1245 early-persist)


marcusquinn:
Resolves #1951.

PR #2015, opened 2026-02-20
t1273: Add supervisor sanity-check for self-healing queue stalls

marcusquinn:
## Summary

- Adds Phase 0.9 sanity check to the supervisor pulse cycle — runs when zero tasks are dispatchable but open tasks exist in TODO.md
- Cross-references DB state, TODO.md state, and system state to find contradictions that cause silent queue stalls
- Fixes Phase 0.5d to treat permanently failed blockers as resolved, preventing infinite dependency chain stalls

## Problem

The supervisor pulse was running every 2 minutes, logging "No new tasks to pick up" hundreds of times, while 13 open tasks sat idle. Three interlocking failures caused the stall:

1. **DB-failed tasks with TODO.md claims (double-lock)**: Tasks failed in DB but still had `assignee:/started:` in TODO.md. Phase 0.5e has a 24h threshold, but the DB already knows the task failed — waiting 24h is unnecessary
2. **Failed dependency chains never cascade**: When a root task fails permanently, its dependents wait forever because Phase 0.5d only checks for `[x]` or `[-]` status
3. **Missing `#auto-dispatch` tag**: Tasks with clear specs, model assignments, and estimates were invisible to auto-pickup because they lacked the tag

### New: `.agents/scripts/supervisor/sanity-check.sh`
Phase 0.9 module with 5 checks:
- **Check 1**: Strip stale claims on DB-failed tasks immediately (DB state is authoritative)
- **Check 2**: Unblock tasks whose blockers permanently failed (retries exhausted)
- **Check 3**: Auto-tag dispatchable tasks missing `#auto-dispatch`
- **Check 4**: Cancel DB orphans with no TODO.md entry
- **Check 5**: Log structured stall breakdown when nothing is fixable

### Modified: `.agents/scripts/supervisor/pulse.sh`
- Integrates Phase 0.9 between Phase 1d (post-completion unblock) and Phase 2 (dispatch)
- Only runs when zero tasks are queued but open tasks exist
- Re-runs auto-pickup after fixes so newly dispatchable tasks enter the queue immediately

### Modified: `.agents/scripts/supervisor/todo-sync.sh`
- Phase 0.5d (`auto_unblock_resolved_tasks`) now checks if a blocker is permanently failed in DB (retries >= max_retries)
- Permanently failed blockers are treated as resolved, unblocking dependents

### Modified: `.agents/scripts/supervisor-helper.sh`
- Sources the new `sanity-check.sh` module

## Design Principle

"Don't assume, verify." Deterministic state machines make assumptions that cause silent failures. When state is contradictory, cross-reference all available data sources to resolve it rather than waiting for timeouts.

## Testing

- ShellCheck: zero violations on all 4 files
- Designed to be triggered by the current stalled queue (0 queued, 0 running, 13 open tasks)

## Summary by CodeRabbit

* **Chores**
  * Supervisor now runs an automated sanity-check phase when the queue is idle.
  * Automatically detects and remediates stale, blocked, or inconsistent task states across repos.
  * Treats permanently failed blockers as resolved to prevent indefinite blocking.
  * Adds automatic tagging for dispatchable-looking tasks missing auto-dispatch.
  * Logs structured stall summaries to improve visibility into why the queue is blocked.

### DIFF
--- a/.agents/scripts/supervisor-helper.sh
+++ b/.agents/scripts/supervisor-helper.sh
@@ -232,6 +232,7 @@ source "${SUPERVISOR_MODULE_DIR}/ai-reason.sh"
 source "${SUPERVISOR_MODULE_DIR}/ai-actions.sh"
 source "${SUPERVISOR_MODULE_DIR}/issue-audit.sh"
 source "${SUPERVISOR_MODULE_DIR}/routine-scheduler.sh"
+source "${SUPERVISOR_MODULE_DIR}/sanity-check.sh"
 
 # Valid states for the state machine
 # shellcheck disable=SC2034 # Used by supervisor/state.sh
--- a/.agents/scripts/supervisor/pulse.sh
+++ b/.agents/scripts/supervisor/pulse.sh
@@ -613,6 +613,31 @@ get_task_timeout() {
 }
 
 #######################################
+# Phase 0.9 helper: run sanity check for a single repo
+# Extracted to avoid duplicating the open-task check + sanity-check + auto-pickup
+# sequence for the multi-repo and single-repo code paths in cmd_pulse.
+#######################################
+_run_sanity_check_for_repo() {
+	local repo_path="$1"
+	local todo_file="$repo_path/TODO.md"
+
+	if [[ ! -f "$todo_file" ]]; then
+		return
+	fi
+
+	local open_count
+	open_count=$(grep -cE '^\s*- \[ \] t[0-9]+' "$todo_file" 2>/dev/null || echo 0)
+	if [[ "$open_count" -gt 0 ]]; then
+		local sanity_fixed
+		sanity_fixed=$(run_sanity_check "$repo_path" 2>>"$SUPERVISOR_LOG")
+		if [[ "${sanity_fixed:-0}" -gt 0 ]]; then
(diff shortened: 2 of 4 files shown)


marcusquinn (commented):
## Review response

### Addressed (valid feedback)
- **Duplicated code in pulse.sh**: Extracted `_run_sanity_check_for_repo()` helper function — good catch
- **Fragile sed for comma-separated blocker removal**: Replaced with robust string rebuild approach (`",$list," | sed "s/,$item,/,/"`)

### Dismissed (incorrect or inconsistent with codebase conventions)

**`2>/dev/null` on SELECT operations**: This is the established convention across the entire supervisor codebase (279 occurrences in pulse.sh alone, 33 in todo-sync.sh, 98 in evaluate.sh). The `db()` wrapper can emit SQLite warnings on locked databases during concurrent access — these are expected in a multi-process supervisor and not actionable errors. Removing `2>/dev/null` would flood the supervisor log with noise during normal operation. The `|| echo ""` fallback handles actual failures.

**`local var="$1"` pattern**: The Gemini suggestion to separate declaration from assignment applies to **command substitutions** where you need to capture exit codes (e.g., `local x; x=$(some_cmd)`). For simple parameter assignments like `local repo_path="$1"`, there is no exit code to capture — `$1` is a variable expansion, not a command. This is the established pattern throughout the codebase (used in every function in todo-sync.sh, pulse.sh, etc.).

Issue #2044, opened 2026-02-21
t1286: Fix claim-task-id.sh --count N to create one GitHub issue per allocated ID


marcusquinn:
**Task ID:** `t1286` | **Status:** open | **Estimate:** `~15m`
**Assignee:** @marcusquinn | **Started:** 2026-02-21
**Tags:** `bugfix`

## Description

Fix claim-task-id.sh --count N to create one GitHub issue per allocated ID

---
*Synced from TODO.md by issue-sync-helper.sh*

marcusquinn:
## Flagged for Human Review

**Reason:** Audit flags no_pr_linkage for t1286, but t1286 is now completed with pr:#2045. Verify the issue was auto-closed by issue-sync after PR merge, then dismiss this audit finding.

This issue has been flagged by the AI supervisor for human review. Please assess and take appropriate action.

---
*Flagged by AI Supervisor (automated reasoning cycle)*

marcusquinn:
## Flagged for Human Review

**Reason:** Audit reports no_pr_linkage for t1286, but task is verified with pr:#2045. Likely a timing issue — audit ran before PR was linked in DB. Verify on next cycle; if persists, the audit script's PR detection may need fixing (related to t1158 which is already open).

This issue has been flagged by the AI supervisor for human review. Please assess and take appropriate action.

---
*Flagged by AI Supervisor (automated reasoning cycle)*

marcusquinn:
## Flagged for Human Review

**Reason:** High-severity audit finding (no_pr_linkage, no_completion_evidence) but t1286 fix has been completed with PR #2045. Issue may need manual closure verification — the fix was to make claim-task-id.sh create one GitHub issue per allocated ID. Verify the fix deployed correctly and close this audit finding.

This issue has been flagged by the AI supervisor for human review. Please assess and take appropriate action.

---
*Flagged by AI Supervisor (automated reasoning cycle)*

marcusquinn:
## Flagged for Human Review

**Reason:** Audit false positive — t1286 was completed with pr:#2045 (merged). The no_pr_linkage finding is stale. Issue can be verified closed once audit DB refreshes.

This issue has been flagged by the AI supervisor for human review. Please assess and take appropriate action.

---
*Flagged by AI Supervisor (automated reasoning cycle)*

marcusquinn:
## Flagged for Human Review

[...]

PR #2045, opened 2026-02-21
t1286: Fix claim-task-id.sh --count N to create one GitHub issue per allocated ID

marcusquinn:
## Summary
- Fixed `claim-task-id.sh --count N` to create one GitHub/GitLab issue per allocated ID instead of only creating one for the first ID
- Output is backwards compatible: `ref=` still points to the first issue
- New `ref_tNNN=GH#NNN` lines added for each issue in batch mode
- Updated help text to document batch issue creation behaviour

## Root Cause
The issue creation block (lines 544-572) was outside any loop — it created exactly one issue using `first_id` regardless of `ALLOC_COUNT`. When `--count 3` allocated t1281-t1283, only t1281 got an issue.

## Fix
Loop issue creation from `first_id` to `last_id`, creating one issue per ID with the same `--title`. Each issue gets its own number tracked in an array, and all refs are output in the machine-readable results.

## Testing
- `shellcheck` clean (only SC1091 for external source)
- `bash -n` syntax check passes
- `--dry-run --count 3` output format verified
- No callers use `--count` programmatically — only interactive and test usage

## Summary by CodeRabbit

* **New Features**
  * Task ID allocation now creates individual issues for each allocated ID in batch operations.

* **Bug Fixes**
  * Improved batch allocation error handling to prevent single-ID failures from stopping the entire batch process.

### DIFF
--- a/.agents/scripts/claim-task-id.sh
+++ b/.agents/scripts/claim-task-id.sh
@@ -10,6 +10,9 @@
 #   --description "Details"    Task description (optional)
 #   --labels "label1,label2"   Comma-separated labels (optional)
 #   --count N                  Allocate N consecutive IDs (default: 1)
+#                              Creates one GitHub/GitLab issue per ID using
+#                              the same --title. Output includes ref_tNNN=GH#NNN
+#                              for each created issue.
 #   --offline                  Force offline mode (skip remote push)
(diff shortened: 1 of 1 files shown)


marcusquinn:
Resolves #2044.

PR #2052, opened 2026-02-21
t1287: Wire -needed blocker tags into cron auto-pickup and AI reasoner park action

marcusquinn:
## Summary

- **cron.sh**: Add `-needed` blocker tag check to `cmd_auto_pickup` (both Strategy 1 and Strategy 2). Tasks tagged with `payment-needed`, `account-needed`, etc. are now skipped during auto-pickup, preventing wasted worker sessions on tasks that require human action.
- **ai-reason.sh**: Add `park_task` as the 11th action type the AI reasoner can propose. Add "Human-action blockers" as the 11th analysis framework step, teaching the AI to identify tasks needing human intervention.
- **ai-actions.sh**: Add `park_task` to valid action types, field validation (validates `blocker_tag` against the 14 known `-needed` tags), and executor (`_exec_park_task`) that appends the tag to the task line in TODO.md and commits.

## Context

The `-needed` blocker tags (t1188.1) were already wired into the eligibility assessment context and sanity-check, but had two gaps:
1. `cron.sh` auto-pickup didn't check for them — a task tagged `#auto-dispatch payment-needed` would still get dispatched
2. The AI reasoner had no way to proactively add `-needed` tags when it identified tasks requiring human action

## Testing

- ShellCheck clean on all three modified files (only pre-existing SC2016/SC1091 info-level)
- Blocker pattern matches the same 14 tags used in `sanity-check.sh` and `ai-context.sh`

### DIFF
--- a/.agents/scripts/supervisor/ai-actions.sh
+++ b/.agents/scripts/supervisor/ai-actions.sh
@@ -41,7 +41,7 @@ AI_SEMANTIC_DEDUP_TIMEOUT="${AI_SEMANTIC_DEDUP_TIMEOUT:-30}"
 AI_IMPROVEMENT_DEDUP_DAYS="${AI_IMPROVEMENT_DEDUP_DAYS:-30}"
 
 # Valid action types — any action not in this list is rejected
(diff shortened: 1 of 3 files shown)


marcusquinn:
Resolves #2050.

PR #2153, opened 2026-02-22
t1309: Intent tracing in OpenCode plugin — log agent__intent alongside tool calls

marcusquinn:
## Summary

- Adds intent tracing to the OpenCode plugin, inspired by oh-my-pi's `agent__intent` pattern
- The LLM is instructed via `experimental.chat.system.transform` to include an `agent__intent` field in every tool call, describing its intent in present participle form
- Intent is extracted in `tool.execute.before`, bridged to `tool.execute.after` via an in-memory Map, and stored in the `tool_calls.intent` column of the observability DB

## Changes

**`observability.mjs`**
- Added `intent TEXT` column to `tool_calls` table schema
- Added migration via `pragma_table_info` check (silent, no noisy errors for existing DBs)
- `recordToolCall()` accepts optional `intent` parameter and stores it in the DB

**`index.mjs`**
- Added `INTENT_FIELD = "agent__intent"` constant (matches oh-my-pi convention)
- Added `intentByCallId` Map to bridge `tool.execute.before` → `tool.execute.after`
- Added `extractAndStoreIntent()` and `consumeIntent()` helpers with memory pruning
- `toolExecuteBefore()`: extracts intent from `output.args`, logs it, stores by callID
- `toolExecuteAfter()`: retrieves intent via `consumeIntent()`, passes to `recordToolCall()`
- `systemTransformHook()`: injects intent instruction into system prompt (always, not gated on TTSR rules)

## Evaluation: tool.execute.before hook

The `tool.execute.before` hook provides **sufficient context** for intent extraction:
- `input.callID` — unique identifier to bridge before→after
- `output.args` — the full tool call arguments including `agent__intent`

The before→after bridge pattern is necessary because `tool.execute.after` doesn't receive args (only `title`, `output`, `metadata`). The in-memory Map approach is consistent with the existing `sessionToolCounts` pattern.

## Why system prompt injection vs JSON Schema injection

OpenCode's plugin API does not expose a hook to modify tool schemas before they are sent to the LLM. The `experimental.chat.system.transform` hook achieves the same chain-of-thought effect as oh-my-pi's JSON Schema injection — the LLM is required to articulate intent before executing each tool.

## Observability query example

```sql
SELECT tool_name, intent, timestamp
FROM tool_calls
WHERE intent IS NOT NULL
ORDER BY timestamp DESC
LIMIT 20;
```

Ref #2133

### DIFF
--- a/.agents/plugins/opencode-aidevops/index.mjs
+++ b/.agents/plugins/opencode-aidevops/index.mjs
@@ -1021,15 +1021,32 @@ function logQualityGateResult(label, filePath, totalViolations, report, errorLev
 }
 
 /**
- * Pre-tool-use hook: Comprehensive quality gate for Write/Edit operations.
- * Runs the full quality pipeline matching pre-commit-hook.sh checks:
- * - Shell scripts (.sh): ShellCheck, return statements, positional params, secrets
- * - Markdown (.md): MD031 (blank lines around code blocks), trailing whitespace
- * - All files: secrets scanning
+ * Pre-tool-use hook: Intent extraction + quality gate for Write/Edit operations.
+ *
+ * Intent tracing (t1309):
+ *   Extracts the `agent__intent` field from tool args (injected by the LLM
+ *   per system prompt instruction) and stores it keyed by callID for
+ *   retrieval in toolExecuteAfter when the call is recorded to the DB.
+ *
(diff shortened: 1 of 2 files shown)


marcusquinn:
Resolves #2190.

PR #2220, opened 2026-02-24
t1317: Migrate routine-scheduler should_run_routine() to AI scheduling

marcusquinn:
## Summary

- Migrate `should_run_routine()` from 120-line case-statement decision tree to AI-driven scheduling via Anthropic API (haiku model)
- Single batch API call evaluates all 5 routines per pulse cycle (cost-efficient: 1 call instead of 5)
- Deterministic heuristic fallback preserved when AI unavailable (no auth, API failure, timeout)
- Hard constraints (min_interval floor, explicit deferrals) always enforced — AI cannot override safety floors
- Auth resolution supports both ANTHROPIC_API_KEY env var and OAuth from auth.json

## Changes

- `_resolve_anthropic_auth()` — bash auth resolution matching TypeScript ai-research pattern
- `_ai_schedule_all_routines()` — batch API call with structured JSON context
- `_heuristic_should_run_routine()` — renamed original case-statement as fallback
- `should_run_routine()` — rewritten to check AI cache first, fall back to heuristic
- `run_phase14_routine_scheduler()` — attempts AI batch evaluation, logs method used

## Testing

- ShellCheck passes clean (zero violations)
- API contract unchanged: callers still get "run"/"skip"/"defer" on stdout, return 0/1
- ROUTINE_DECISION_* env var exports unchanged — pulse.sh consumers unaffected

Ref #2178

## Release Notes

* **New Features**
  * Routine scheduling now leverages AI-powered decision-making for optimized routine execution.

* **Improvements**
  * Added automatic fallback to deterministic logic when AI services are unavailable, ensuring scheduling continuity.
  * Enhanced logging to indicate which scheduling method was applied to each routine decision.

### DIFF
--- a/.agents/scripts/supervisor/routine-scheduler.sh
+++ b/.agents/scripts/supervisor/routine-scheduler.sh
@@ -1,18 +1,25 @@
 #!/usr/bin/env bash
-# routine-scheduler.sh - Intelligent routine scheduling for supervisor pulse (t1093)
+# routine-scheduler.sh - AI-driven routine scheduling for supervisor pulse (t1093, t1317)
 #
 # Phase 14: AI-driven routine scheduling that adapts frequency and priority
-# based on project state signals. Instead of hardcoded daily/hourly schedules,
-# evaluates whether each routine is worth running now.
+# based on project state signals. Uses Anthropic API (haiku) to make scheduling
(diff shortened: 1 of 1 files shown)


marcusquinn:
Resolves #2213.

PR #2309, opened 2026-02-25
t1327.2: Enhance SimpleX Chat subagent with bot API reference details

marcusquinn:
## Summary

- Enhance the SimpleX Chat subagent documentation with detailed bot API reference material from the t1327.1 research report
- Add essential bot events table (9 events with type tags, triggers, and recommended actions)
- Add ChatRef syntax reference (`@<contactId>`, `#<groupId>`, `*<noteFolderId>`)
- Add network usage classification table (no/interactive/background)
- Expand TypeScript SDK section with full `ChatClient` API examples (connect, get user, send messages, event loop, raw commands)
- Add key types reference (MsgContent, ComposedMessage, ChatBotCommand, AddressSettings, GroupMemberRole)
- Add Bun compatibility notes for bot development

These additions make the doc self-contained for bot developers — common operations no longer require cross-referencing upstream API docs.

## Changes

| Section | What was added |
|---------|---------------|
| Essential Bot Events | 9-row table: contactConnected, acceptingBusinessRequest, newChatItems, receivedContactRequest, rcvFileDescrReady, rcvFileComplete, receivedGroupInvitation, joinedGroupMember, deletedMemberUser |
| ChatRef Syntax | 3-row table mapping prefixes to chat targets |
| Network Usage Classification | 3-row table explaining no/interactive/background command behaviour |
| TypeScript SDK | Full ChatClient code example (create, get user, address, auto-accept, send, event loop, raw command) + Bun compatibility note |
| Key Types | MsgContent union, ComposedMessage, ChatBotCommand, AddressSettings, GroupMemberRole |

Ref #2247

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive TypeScript SDK usage examples for connection, message handling, event loop, and sample responses
  * Introduced Essential Bot Events reference table (repositioned for clarity)
  * Added ChatRef syntax guide with prefix conventions and examples
  * Documented Network Usage Classification (no, interactive, background) with examples
  * Documented Key Types for message content, composed messages, commands, and address settings
  * Added Bun compatibility guidance for session storage and WebSocket notes

PR #2266, opened 2026-02-25
t1327.2: SimpleX Chat subagent documentation

marcusquinn:
## Summary

- Create comprehensive SimpleX Chat subagent documentation at `.agents/services/communications/simplex.md`
- Covers all required sections: install, bot API, business addresses, protocol, voice/files, multi-platform, cross-device workarounds, self-hosted servers, upstream contributions, limitations
- Follows existing conventions from `matrix-bot.md` and `twilio.md` (YAML frontmatter, AI-CONTEXT markers, section structure)

## Sections

| Section | Content |
|---------|---------|
| Quick Reference | Type, license, apps, SDK, install command, encryption, key differentiator, SimpleX vs Matrix comparison |
| Architecture | Diagram showing app → SMP relay → CLI → bot process flow |
| Installation | CLI (Linux/macOS/Windows), mobile/desktop apps, build from source |
| CLI Usage | Running modes, essential commands table, database details |
| Bot API | WebSocket server, command/event JSON format, key commands, TypeScript SDK, bot profile, command menus, constraints |
| Business Addresses | How they work, use cases, setup instructions |
| Protocol Overview | SMP, XFTP, WebRTC details, encryption primitives, comparison with Signal |
| Voice/Files/Calls | Voice messages, file transfer, audio/video calls |
| Multi-Platform | All platforms with features and notes, incognito mode, multiple profiles |
| Cross-Device Workarounds | Remote Control Protocol (XRCP), cloud CLI + tmux, database migration |
| Self-Hosted Servers | SMP server, XFTP server, WebRTC TURN/STUN, app configuration |
| Upstream Contributions | Repo structure, contributing guide, feedback channels, logging template |
| Limitations | Cross-device sync, single profile, owner role recovery, group stability, no search, API auth, file limits, push notifications |
| Matterbridge | Bridge to 40+ platforms, privacy gradient |
| Security | Threat model, bot security model, operational security |
| aidevops Integration | Planned components table, slash command coexistence |

## Task

t1327.2: Subagent documentation

Ref #2247

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive SimpleX Chat integration guide covering system architecture, installation procedures, bot API usage, protocol specifications, security considerations, cross-device synchronization, self-hosted server deployment, multi-profile management, and operational guidelines for users and developers.

PR #2317, opened 2026-02-25
t1327.2: Fix SimpleX subagent documentation review feedback

marcusquinn:
## Summary

- Fix `ChatType` import to use correct module path (`simplex-chat/dist/command` instead of `@simplex-chat/types`)
- Fix `enableAddressAutoAccept()` call signature (remove incorrect `userId` parameter)
- Fix hardcoded `1` in `sendChatCmd` raw command example to use dynamic `user.userId`
- Add placeholder type definitions for `LinkPreview` and `CryptoFile` to make Key Types section self-contained

## Context

Addresses unresolved review comments from PR #2266 (CodeRabbit, Augment, Gemini) and PR #2309 (CodeRabbit critical + Augment medium findings) that were merged without all feedback being addressed.

Ref #2247

### DIFF
--- a/.agents/services/communications/simplex.md
+++ b/.agents/services/communications/simplex.md
@@ -330,7 +330,7 @@ The SDK provides a typed WebSocket client with sequential command queue and type
 
 ```typescript
 import {ChatClient} from "simplex-chat"
-import {ChatType} from "@simplex-chat/types"
+import {ChatType} from "simplex-chat/dist/command"
 
 // Connect to CLI WebSocket server
 const chat = await ChatClient.create("ws://localhost:5225")
@@ -343,7 +343,7 @@ const address = await chat.apiGetUserAddress(user.userId)
   || await chat.apiCreateUserAddress(user.userId)
 
 // Enable auto-accept for incoming contacts
-await chat.enableAddressAutoAccept(user.userId)
+await chat.enableAddressAutoAccept()
 
 // Send text message to contact or group (use ChatType enum)
 const contactId = 1  // obtained from contactConnected event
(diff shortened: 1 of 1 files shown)


marcusquinn (commented):
## Review: Changes Verified Against Upstream Source

I verified all 4 changes against the [upstream `client.ts`](https://github.com/simplex-chat/simplex-chat/blob/stable/packages/simplex-chat-client/typescript/src/client.ts) on the `stable` branch. **2 of 4 changes are incorrect and need fixing before merge.**

### 1. `ChatType` import path — INCORRECT

The PR changes from `@simplex-chat/types` to `simplex-chat/dist/command`. However, the upstream source shows:

```typescript
// client.ts imports types as:
import {CC, ChatResponse, CR, ChatEvent, T} from "@simplex-chat/types"
```

And `index.ts` only exports `ChatClient` — no types are re-exported. The `simplex-chat/dist/command` path does not exist in the published package. `ChatType` lives in `@simplex-chat/types` as `T.ChatType`.

**Fix**: Revert to `@simplex-chat/types` but show the correct usage pattern — either `import {T} from "@simplex-chat/types"` then `T.ChatType`, or if the types package has a direct export, use that. The Augment reviewer was correct on this one.

### 2. `enableAddressAutoAccept()` — INCORRECT (userId removal)

The upstream signature is:

```typescript
async enableAddressAutoAccept(userId: number, autoReply?: T.MsgContent, businessAddress = false): Promise<void>
```

`userId` is **required**. The PR removes it, which would cause a runtime error. The Augment reviewer was correct.

**Fix**: Restore `user.userId` as the first argument: `await chat.enableAddressAutoAccept(user.userId)`

### 3. Dynamic `user.userId` in `sendChatCmd` — CORRECT

Replacing hardcoded `1` with `${user.userId}` is the right fix.

### 4. Placeholder types for `LinkPreview` and `CryptoFile` — CORRECT

These are reasonable additions. Consider Gemini's suggestion to use `Record<string, any>` instead of `object` for `cryptoArgs`.

### Summary

Please fix items 1 and 2 before merging. The original code was actually correct for the import path and function signature — the previous review feedback that prompted these changes was itself incorrect (AI reviewers hallucinated the issues). Only items 3 and 4 should be kept.

PR #2853, opened 2026-03-05
t1397: fix pulse-wrapper stuck process — add watchdog timeout

marcusquinn:
## Summary

- Replace bare `wait $opencode_pid` in `run_pulse()` with a watchdog loop that polls every 60s and kills the process tree if it exceeds `PULSE_STALE_THRESHOLD` (30min)
- Simplify quality sweep timestamp validation — strip non-numeric chars and default to 0

## Root Cause

The `marcusquinn` pulse runner was stuck for 28+ hours. The design assumed launchd `StartInterval` would fire new invocations every 120s, and `check_dedup()` in the new invocation would detect and kill stale processes. But launchd `StartInterval` only fires when the **previous invocation has exited** — and the wrapper was blocked on `wait $opencode_pid`, so it never exited, and no new invocation ever started.

The `opencode run` process hung after starting the pulse workflow — the LLM session started but never produced output or completed. With no watchdog, the wrapper waited indefinitely.

## Fix

The watchdog is now **internal** to `run_pulse()` — the same process that spawned opencode checks every 60s whether the child has exceeded the stale threshold. If so, it kills the process tree (SIGTERM then SIGKILL) and continues to the quality sweep and health issue phases.

`check_dedup()` remains as a secondary safety net for edge cases where the wrapper itself gets stuck.

Portable across bash 3.2+ (macOS default) — no `wait -n -t` dependency.

## Testing

- ShellCheck clean
- Logic verified: the `while kill -0` loop exits when the process dies naturally OR when elapsed > threshold
- The stuck processes (PIDs 56504/56743/56744) need manual kill after merge + deploy

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of stalled processes to prevent indefinite hangs
  * Enhanced process monitoring and automatic cleanup for better system stability
  * Fixed timestamp handling to be more robust and reliable

### DIFF
--- a/.agents/scripts/pulse-wrapper.sh
+++ b/.agents/scripts/pulse-wrapper.sh
@@ -8,13 +8,14 @@
 #   1. Uses a PID file with staleness check (not pgrep) for dedup
 #   2. Cleans up orphaned opencode processes before each pulse
 #   3. Calculates dynamic worker concurrency from available RAM
-#   4. Lets the pulse run to completion — no hard timeout
+#   4. Internal watchdog kills stuck pulses after PULSE_STALE_THRESHOLD (t1397)
 #
 # Lifecycle: launchd fires every 120s. If a pulse is still running, the
-# dedup check skips. If a pulse has been running longer than PULSE_STALE_THRESHOLD
-# (default 30 min), it's assumed stuck (opencode idle bug) and killed so the
-# next invocation can start fresh. This is the ONLY kill mechanism — no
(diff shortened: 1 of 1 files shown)


alex-solovyev:
Pulse triage: PR is mergeable (author: marcusquinn, maintainer). CodeRabbit review is in-progress — will merge once CodeRabbit posts its final review. All other checks pass (SonarCloud, Codacy, qlty, Socket).

PR #2909, opened 2026-03-05
t1403: fix macOS timeout compatibility and UTC timezone parsing in aidevops update

### LINKED ISSUE #2490: t1356: perf: fix aidevops update exceeding 120s timeout
**Task ID:** `t1356` | **Status:** open
**Assignee:** @alex-solovyev
**Tags:** `bugfix` `performance`

## Description

perf: fix aidevops update exceeding 120s timeout — P0: add integrity check after generate-opencode-agents.sh to detect incomplete subagent generation and print recovery command. P1: use skill-scanner scan-all batch mode instead of serial per-file invocations in security-helper.sh (removes

---
*Synced from TODO.md by issue-sync-helper.sh*

johnwaldo:
## Summary

- **macOS has no native `timeout` command** — `aidevops update` calls `timeout 30 npm view ...` and `timeout 120 bash -c "$update_cmd"` which silently fail (exit 127) or hang on macOS. Added `_timeout_cmd()` in `aidevops.sh` and `timeout_sec()` in `tool-version-check.sh` using perl alarm fallback (available on all macOS).
- **Pipe to `tail` hangs with perl alarm** — even with `timeout_sec` working, `timeout_sec 120 bash -c "$cmd" 2>&1 | tail -2` hangs forever on macOS. The perl alarm (`SIGALRM`) kills the process but doesn't close the pipe's write end, so `tail` blocks waiting for EOF. Replaced with temp file: write output to file, then `tail -2` the file.
- **UTC timezone parsing bug** — `date -j -f` on macOS ignores the `Z` suffix, interpreting UTC timestamps in local time. This caused freshness checks to be off by the local UTC offset (e.g., 7 hours for MST), potentially skipping update checks. Fixed by prefixing with `TZ=UTC`.

## Files Changed

- `aidevops.sh` — Added `_timeout_cmd()` portable function, replaced 2 bare `timeout` calls
- `.agents/scripts/tool-version-check.sh` — Added `timeout_sec()` portable function, replaced bare `timeout 120` in update execution, replaced `| tail -2` pipe with temp file to prevent SIGALRM hang
- `.agents/scripts/auto-update-helper.sh` — Fixed 3 `date -j` calls to use `TZ=UTC`

## Root Cause Analysis

Two separate bugs causing the same symptom (hang at tool update):

1. **Missing `timeout` command** — macOS has no `timeout`. The shell tries to execute it, fails with exit 127, and `|| true` swallows the error. Fixed with `timeout_sec()` / `_timeout_cmd()` using `perl -e 'alarm shift; exec @ARGV'`.

2. **SIGALRM + pipe deadlock** — Even with the perl alarm working, piping its output to `tail -2` creates a deadlock. When SIGALRM fires, perl sends the signal to the exec'd process, but the pipe's write-end file descriptor isn't closed cleanly. `tail` holds the read end open, waiting for EOF that never comes. Linux's `timeout` command avoids this by using process groups (`kill -TERM -- -$pgid`). Fixed by redirecting to a temp file instead of piping.

## Testing

Verified on macOS (no native `timeout`):
- `_timeout_cmd` correctly times out via perl alarm (exit 142)
- `timeout_sec` correctly wraps npm/brew queries
- UTC fix eliminates 7-hour offset in freshness calculations
- Temp file approach correctly shows last 2 lines of output and terminates cleanly

Closes #2490

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed timestamp parsing on macOS so UTC-formatted timestamps (trailing Z) are handled correctly.

* **Chores**
  * Standardized timeout handling across scripts with a portable wrapper to prevent hangs.
  * Capture update/version output to temporary logs and surface concise end-of-operation output (last lines) for clearer success/failure feedback.

### DIFF
--- a/.agents/scripts/auto-update-helper.sh
+++ b/.agents/scripts/auto-update-helper.sh
@@ -414,7 +414,8 @@ check_skill_freshness() {
 	if [[ -n "$last_skill_check" ]]; then
 		local last_epoch now_epoch elapsed
 		if [[ "$(uname)" == "Darwin" ]]; then
-			last_epoch=$(date -j -f "%Y-%m-%dT%H:%M:%SZ" "$last_skill_check" "+%s" 2>/dev/null || echo "0")
+			# TZ=UTC: stored timestamps are UTC — macOS date -j ignores the Z suffix
+			last_epoch=$(TZ=UTC date -j -f "%Y-%m-%dT%H:%M:%SZ" "$last_skill_check" "+%s" 2>/dev/null || echo "0")
 		else
 			last_epoch=$(date -d "$last_skill_check" "+%s" 2>/dev/null || echo "0")
 		fi
@@ -533,7 +534,8 @@ check_openclaw_freshness() {
 	if [[ -n "$last_openclaw_check" ]]; then
 		local last_epoch now_epoch elapsed
 		if [[ "$(uname)" == "Darwin" ]]; then
-			last_epoch=$(date -j -f "%Y-%m-%dT%H:%M:%SZ" "$last_openclaw_check" "+%s" 2>/dev/null || echo "0")
+			# TZ=UTC: stored timestamps are UTC — macOS date -j ignores the Z suffix
+			last_epoch=$(TZ=UTC date -j -f "%Y-%m-%dT%H:%M:%SZ" "$last_openclaw_check" "+%s" 2>/dev/null || echo "0")
 		else
 			last_epoch=$(date -d "$last_openclaw_check" "+%s" 2>/dev/null || echo "0")
 		fi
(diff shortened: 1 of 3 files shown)


marcusquinn:
This PR is from an external contributor (@johnwaldo). Auto-merge is disabled for external PRs — a maintainer must review and merge manually.

robstiles:
Permission check failed for this PR (HTTP 403 from collaborator permission API). Unable to determine if @johnwaldo is a maintainer or external contributor. **A maintainer must review and merge this PR manually.** This is a fail-closed safety measure — the pulse will not auto-merge until the permission API succeeds.

PR #2922, opened 2026-03-05
fix: add timeout to get_installed_version() in tool-version-check.sh

### LINKED ISSUE #2920: bug: tool-version-check.sh hangs — get_installed_version has no timeout on local --version calls
## Summary

`tool-version-check.sh` hangs indefinitely when a tool's `--version` command doesn't exit, blocking `aidevops update` and eventually causing system resource exhaustion.

## Root Cause

`get_installed_version()` (line 116) runs `$cmd $ver_flag` with **no timeout**:

```bash
version=$("$cmd" $ver_flag 2>/dev/null | head -1 | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' | head -1 || echo "")
```

The script already has a `timeout_sec()` function and a `PKG_QUERY_TIMEOUT=30` constant, but these are only applied to **remote** queries (`npm view`, `brew info`, `pip index`). The local `--version` calls have no timeout at all.

## Observed Failure

- `macos-automator-mcp --version` (PID 30884) hung indefinitely — the node process started but never exited
- This blocked three parent `tool-version-check.sh` processes (PIDs 30790, 30882, 30883)
- Which blocked the parent `aidevops.sh update` (PID 2752)
- All five processes were stuck in `S+` state on terminal s007
- Required manual `kill` to recover

## Impact

- **Blocks `aidevops update` entirely** — no tools get checked/updated
- **Terminal session locked** — the `S+` (foreground stopped) state means the terminal is unusable
- **System stability risk** — if this happens during automated/scheduled runs, hung processes accumulate

## Proposed Fix

Wrap the `--version` call in `timeout_sec` with a short timeout (5-10 seconds). A local `--version` command should never take more than a few seconds:

```bash
get_installed_version() {
    local cmd="$1"
    local ver_flag="$2"
    local VERSION_TIMEOUT=10

    if command -v "$cmd" &>/dev/null; then
        local version
        # shellcheck disable=SC2086
        version=$(timeout_sec "$VERSION_TIMEOUT" "$cmd" $ver_flag 2>/dev/null | head -1 | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' | head -1 || echo "")
        if [[ -z "$version" ]]; then
            # shellcheck disable=SC2086
            version=$(timeout_sec "$VERSION_TIMEOUT" "$cmd" $ver_flag 2>/dev/null | head -1 | grep -oE '[0-9]+\.[0-9]+' | head -1 || echo "unknown")
        fi
        echo "$version"
    else
        echo "not installed"
    fi
}
```

Also consider: if `timeout_sec` kills the process, return `"timeout"` instead of `"unknown"` so the output distinguishes "couldn't determine version" from "command hung".

## Environment

- macOS (no native `timeout` — uses `perl -e 'alarm shift; exec @ARGV'` fallback)
- Hung tool: `@steipete/macos-automator-mcp` (npm global)
- aidevops installed via Homebrew at `/opt/homebrew/lib/node_modules/aidevops/`

## Affected File

`agents/scripts/tool-version-check.sh`, lines 109-125 (`get_installed_version` function)

robstiles:
## Summary

- Wraps local `--version` calls in `timeout_sec` with a 10-second limit to prevent indefinite hangs when tools (e.g. `macos-automator-mcp`) don't exit
- Returns `"timeout"` instead of `"unknown"` so output distinguishes "command hung" from "couldn't parse version"
- Captures raw output once then parses, avoiding running the command twice

## Details

`get_installed_version()` ran `$cmd $ver_flag` with no timeout. The script already had `timeout_sec()` and `PKG_QUERY_TIMEOUT=30` for remote queries, but local `--version` calls were unprotected. This caused `macos-automator-mcp --version` to hang indefinitely, blocking `aidevops update` and locking the terminal.

### Changes

- Add `VERSION_TIMEOUT=10` constant (local commands should respond in <1s; 10s avoids false positives)
- Wrap the `--version` call in `timeout_sec "$VERSION_TIMEOUT"`
- Detect timeout via exit codes 124 (GNU `timeout`/`gtimeout`) and 142 (perl `alarm` = 128+SIGALRM)
- Add `"timeout"` status handling in `check_tool()` with distinct icon and message
- Skip remote latest-version lookup when local version timed out (pointless)

### Testing

- `bash -n` syntax check passes
- Timeout exit code handling covers all three `timeout_sec` backends: GNU `timeout` (124), `gtimeout` (124), perl `alarm` (142)

Closes #2920

## Summary by CodeRabbit

* **Bug Fixes**
  * Added timeout protection to tool version commands to prevent indefinite hangs when version detection is slow or unresponsive.
  * Improved version detection reliability with fallback parsing patterns.
  * Enhanced output to clearly report when version commands exceed the timeout threshold.

### DIFF
--- a/.agents/scripts/tool-version-check.sh
+++ b/.agents/scripts/tool-version-check.sh
@@ -105,6 +105,10 @@ NOT_INSTALLED_COUNT=0
 declare -a OUTDATED_PACKAGES=()
 declare -a JSON_RESULTS=()
 
+# Timeout for local --version commands (seconds)
+# Local commands should respond in <1s; 10s is generous to avoid false positives
+readonly VERSION_TIMEOUT=10
+
 # Get installed version
 get_installed_version() {
 	local cmd="$1"
@@ -112,12 +116,19 @@ get_installed_version() {
 
 	if command -v "$cmd" &>/dev/null; then
 		local version
+		local raw_output
+		local exit_code=0
 		# shellcheck disable=SC2086
-		version=$("$cmd" $ver_flag 2>/dev/null | head -1 | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' | head -1 || echo "")
+		raw_output=$(timeout_sec "$VERSION_TIMEOUT" "$cmd" $ver_flag 2>/dev/null) || exit_code=$?
+		# Exit 124 = GNU timeout/gtimeout, 142 = perl alarm (128+SIGALRM)
+		if [[ $exit_code -eq 124 || $exit_code -eq 142 ]]; then
+			echo "timeout"
+			return 0
+		fi
+		version=$(echo "$raw_output" | head -1 | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' | head -1 || echo "")
 		if [[ -z "$version" ]]; then
 			# Try alternative patterns
-			# shellcheck disable=SC2086
-			version=$("$cmd" $ver_flag 2>/dev/null | head -1 | grep -oE '[0-9]+\.[0-9]+' | head -1 || echo "unknown")
+			version=$(echo "$raw_output" | head -1 | grep -oE '[0-9]+\.[0-9]+' | head -1 || echo "unknown")
 		fi
 		echo "$version"
 	else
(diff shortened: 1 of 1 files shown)


robstiles:
Closing: created by automated supervisor pulse without user knowledge.

PR #2924, opened 2026-03-05
t2899: Fix security findings in memory-pressure-monitor.sh

robstiles:
## Summary

- **HIGH**: Fix AppleScript injection in `notify()` — replace string interpolation with stdin script + `argv` passing, eliminating the attack vector where attacker-controlled process names could execute arbitrary commands via `"; do shell script "...`
- **MEDIUM**: Validate `MEMORY_COOLDOWN_SECS` with `_validate_int()` to prevent command injection via bash arithmetic expansion
- **MEDIUM**: Remove redundant `2>/dev/null` from `cat` after file existence check, per project conventions
- **MEDIUM**: Optimize process collection and session counting loops — replace `echo|awk|cut` subshells with `read` builtin and parameter expansion

## Security Fix Detail

The `notify()` function previously escaped `\` and `"` in title/message before interpolating them into an `osascript -e` string. This was insufficient — an attacker could break out of the AppleScript string literal with a payload like `"; do shell script "whoami`. Since process names (which appear in notification messages) can be controlled by any user on the system, this was exploitable.

**Fix**: Pass the AppleScript via stdin using `on run argv` handler. Title and message are passed as positional arguments to `osascript`, never appearing in the script text. No escaping needed, no breakout possible.

## Testing

- 31 tests pass (28 original + 3 new)
- ShellCheck: zero violations on both files
- New tests: `MEMORY_COOLDOWN_SECS` injection, osascript stdin pattern verification, AppleScript injection payload handling

```
Results: 31 passed (total: 31)
```

Closes #2899


robstiles:
Closing: created by automated supervisor pulse without user knowledge.

PR #2934, opened 2026-03-05
t2929: Fix config-helper.sh _jsonc_get discards false values

### LINKED ISSUE #2929: bug: config-helper.sh _jsonc_get discards false values — boolean configs cannot be set to false
## Summary

The `_jsonc_get()` function in `config-helper.sh` uses `jq` with the `// empty` alternative operator, which treats `false` as falsy. This means any boolean config set to `false` in `config.jsonc` is silently discarded, and the default value (typically `true`) is returned instead.

**You cannot disable any feature via the JSONC config system.** The only working override is environment variables.

## Root cause

`config-helper.sh` line in `_jsonc_get()`:

```bash
value=$(echo "$merged" | jq -r --arg p "$dotpath" 'getpath($p | split(".")) // empty' 2>/dev/null)
```

In jq, `//` is the alternative operator. `false // empty` returns `empty` because `false` is falsy in jq. The empty string then triggers the default fallback in the bash code.

## Reproduction

```bash
# Set supervisor_pulse to false
~/.aidevops/agents/scripts/config-helper.sh set orchestration.supervisor_pulse false

# Verify the file is correct
cat ~/.config/aidevops/config.jsonc
# Shows: "supervisor_pulse": false  ← correct

# Verify merge is correct
source ~/.aidevops/agents/scripts/config-helper.sh
_get_merged_config | jq .orchestration
# Shows: "supervisor_pulse": false  ← correct

# But config_get returns the default instead
config_get orchestration.supervisor_pulse "UNSET"
# Returns: UNSET  ← wrong, should return "false"

# So is_feature_enabled falls through to the default (true)
is_feature_enabled supervisor_pulse && echo ENABLED || echo DISABLED
# Returns: ENABLED  ← wrong
```

## Fix

Replace `// empty` with a null-only check. For example:

```jq
getpath($p | split(".")) | if . == null then empty else . end
```

Or handle it in the bash wrapper:

```bash
value=$(echo "$merged" | jq --arg p "$dotpath" 'getpath($p | split("."))' 2>/dev/null)
if [[ "$value" == "null" || -z "$value" ]]; then
    echo "$default"
else
    echo "$value" | jq -r '.'
fi
```

## Impact

Every boolean config that defaults to `true` cannot be set to `false` via the config file. This affects `supervisor_pulse`, `auto_update`, `repo_sync`, `hooks_enabled`, and every other boolean toggle in the defaults. The env var override path (`AIDEVOPS_SUPERVISOR_PULSE=false`) works because it bypasses `_jsonc_get()` entirely.

Discovered while investigating #2926 — user set `supervisor_pulse: false` in config.jsonc but the pulse block in setup.sh still activated.

marcusquinn:
## Summary

- Replace jq `// empty` alternative operator with explicit null checks in bash across 5 locations in `config-helper.sh`
- `false` and `0` values are now correctly preserved instead of being silently discarded and replaced with defaults
- Boolean configs (e.g., `supervisor_pulse: false`) can now be disabled via `config.jsonc` as intended

## Root Cause

jq's `//` (alternative) operator treats both `null` and `false` as falsy. So `false // empty` returns `empty`, which becomes an empty string in bash, triggering the default fallback. This affected `_jsonc_get`, `_jsonc_get_raw`, `cmd_list`, and `cmd_set`.

## Fix

Get the raw JSON value (without `-r` or `// empty`), check for literal `"null"` string in bash, then render with `jq -r '.'` only for non-null values. This correctly distinguishes `null` (key absent) from `false`/`0` (key present with falsy value).

## Verification

Tested all paths: `_jsonc_get`, `_jsonc_get_raw`, `config_get`, `config_enabled`, `is_feature_enabled` — all correctly return `false` for false values and fall through to defaults only for absent keys. ShellCheck clean (no new violations).

Closes #2929

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration value handling to correctly distinguish between absent settings and explicit falsy values (false, 0), enhancing reliability of configuration retrieval and defaults.

### DIFF
--- a/.agents/scripts/config-helper.sh
+++ b/.agents/scripts/config-helper.sh
@@ -233,13 +233,17 @@ _jsonc_get() {
 	merged=$(_get_merged_config)
 
 	# Use jq --arg to safely pass dotpath (no shell interpolation into filter)
-	local value
-	value=$(echo "$merged" | jq -r --arg p "$dotpath" 'getpath($p | split(".")) // empty' 2>/dev/null) || value=""
+	# NOTE: We must NOT use jq's `// empty` alternative operator here because
+	# it treats `false` and `null` the same (both are falsy in jq). Instead,
+	# get the raw JSON value and check for null explicitly in bash.
+	local raw_value
+	raw_value=$(echo "$merged" | jq --arg p "$dotpath" 'getpath($p | split("."))' 2>/dev/null) || raw_value="null"
 
-	if [[ -n "$value" && "$value" != "null" ]]; then
-		echo "$value"
-	else
+	if [[ "$raw_value" == "null" ]]; then
 		echo "$default"
+	else
+		# Use jq -r to strip quotes from strings; booleans/numbers pass through as-is
+		echo "$raw_value" | jq -r '.'
 	fi
 	return 0
 }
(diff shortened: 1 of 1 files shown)


marcusquinn:
Closing — this PR fixes the same bug as #2931 which was merged first. Issue #2929 is now resolved. PR #2934 has merge conflicts as a result of #2931 landing.

PR #2918, opened 2026-03-05
GH#2915: fix ShellCheck memory explosion from language server (11 GB RSS, kernel panics)

marcusquinn:
## Summary

- **Root cause**: bash language server hardcodes `--external-sources` in every ShellCheck invocation. With 463 cross-sourcing scripts, this causes exponential AST expansion (11 GB RSS, kernel panics)
- **Fix**: ShellCheck wrapper script strips `--external-sources`; memory monitor auto-kills runaway processes; setup.sh deploys and configures automatically
- **Activation**: `setup.sh` / `aidevops update` sets `SHELLCHECK_PATH` via `launchctl setenv` (macOS GUI) + shell rc files (all terminals)

## Changes

| File | What |
|------|------|
| `.agents/scripts/shellcheck-wrapper.sh` | **Root cause fix** — strips `--external-sources`, enforces ulimit, finds real shellcheck on PATH |
| `.agents/scripts/.shellcheckrc` | Secondary defense — `external-sources=false` for direct file invocations |
| `.agents/scripts/memory-pressure-monitor.sh` | Safety net — auto-kills ShellCheck at >4 GB RSS or >10 min runtime |
| `setup-modules/shell-env.sh` | New `setup_shellcheck_wrapper()` — configures SHELLCHECK_PATH cross-shell |
| `setup.sh` | Calls `setup_shellcheck_wrapper()` in both interactive and non-interactive modes |

## Verification

- Wrapper tested: exact language server invocation pattern completes instantly (was timing out at 10s+)
- ShellCheck clean on all modified scripts
- All 28 memory-pressure-monitor tests pass
- `--status` and `--help` output updated with auto-kill documentation

## Post-merge

After merge, run `aidevops update` (or `./setup.sh`) to deploy the wrapper and set `SHELLCHECK_PATH`. Restart terminal/OpenCode for the env var to take effect.

Closes #2915

PR #2937, opened 2026-03-05
fix: write SHELLCHECK_PATH to .zshenv for non-interactive shell coverage

marcusquinn:
## Summary

- Fixes the recurring shellcheck memory explosion (17.8 GB across 73 processes today) by writing `SHELLCHECK_PATH` to `~/.zshenv` in addition to `.zshrc`
- `.zshenv` is sourced by ALL zsh processes (interactive and non-interactive), ensuring the shellcheck-wrapper is visible to OpenCode's bash-language-server child processes

## Problem

PR #2918 (GH#2915) added `shellcheck-wrapper.sh` and configured `SHELLCHECK_PATH` in `.zshrc` and `.bash_profile`. But `.zshrc` is only sourced for **interactive** shells. OpenCode spawns `bash-language-server` as a non-interactive child process, so `SHELLCHECK_PATH` was never set — the wrapper was completely bypassed.

The bash-language-server hardcodes `--external-sources` in every shellcheck invocation. Without the wrapper stripping this flag, shellcheck follows all `source` directives recursively across 460+ scripts, causing exponential memory growth per invocation.

## Fix

Add `.zshenv` as Layer 2 in `setup_shellcheck_wrapper()`:
1. **launchctl setenv** (macOS) — GUI-launched apps (current boot only)
2. **`.zshenv`** — ALL zsh processes including non-interactive (persists across reboots) **[NEW]**
3. **Shell rc files** (.zshrc, .bash_profile) — interactive terminals

## Related

- Closes the gap left by PR #2918
- Filed upstream: https://github.com/anomalyco/opencode/issues/16209

### DIFF
--- a/setup-modules/shell-env.sh
+++ b/setup-modules/shell-env.sh
@@ -463,8 +463,13 @@ add_local_bin_to_path() {
 # in every ShellCheck invocation, causing exponential memory growth (11 GB+)
 # when source chains span 463+ scripts. The wrapper intercepts this.
 #
-# Uses launchctl setenv (macOS) for GUI-launched apps + shell rc for terminals.
-# This ensures all processes — regardless of shell — see the wrapper.
+# Three layers ensure all processes see the wrapper:
+#   1. launchctl setenv (macOS) — GUI-launched apps (current boot only)
+#   2. .zshenv — ALL zsh processes including non-interactive (persists)
+#   3. Shell rc files (.zshrc, .bash_profile) — interactive terminals
+# Layer 2 is critical: opencode spawns bash-language-server as a non-interactive
+# child process. .zshrc is NOT sourced for non-interactive shells, so without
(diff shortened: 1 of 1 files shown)


marcusquinn:
@coderabbitai review

marcusquinn:
**Follow-up: root cause removed in PR #2939**

This PR fixed the .zshenv path for non-interactive shells. PR #2939 goes further — removes `source-path=SCRIPTDIR` from `.shellcheckrc` entirely and disables SC1091 globally. Even if the wrapper is bypassed, there's nothing to explode now.

**Upstream hazard note:** The bash-language-server still hardcodes `--external-sources` — other opencode/shellcheck users with `source-path` in their `.shellcheckrc` will hit the same issue. See [anomalyco/opencode#16209](https://github.com/anomalyco/opencode/issues/16209).

PR #2885, opened 2026-03-05
t1398.2: Harden ShellCheck invocation to prevent exponential expansion

alex-solovyev:
## Summary

- Harden ShellCheck invocations in `pulse-wrapper.sh` and `linters-local.sh` to prevent the exponential source expansion that caused the March 3 kernel panic (5.7 GB RSS, 88% CPU, 35+ min runtime)
- Add `--norc` flag to pulse-wrapper.sh quality sweep to ignore `.shellcheckrc`'s `source-path=SCRIPTDIR` directive
- Add `ulimit -v 1GB` per-subprocess memory cap as defense-in-depth in both scripts

### pulse-wrapper.sh (quality sweep)
- **`--norc`**: Prevents `.shellcheckrc` from being read, which sets `source-path=SCRIPTDIR` and could cause implicit source following
- **No `-x`/`--external-sources`**: Already the case, now explicitly documented
- **`ulimit -v 1048576`**: Caps virtual memory to 1 GB per shellcheck subprocess
- **Updated `guard_child_processes()` comment**: Clarifies this is a secondary defense, not primary

### linters-local.sh (interactive linting)
- **`ulimit -v 1048576`**: Added to both timeout and portable-fallback paths
- **Keeps `-x -P SCRIPTDIR`**: Interactive use benefits from source resolution; per-file timeout + ulimit provide sufficient protection
- **Updated comments**: Documents the defense-in-depth strategy

### .shellcheckrc
- **Warning comment**: Documents the exponential expansion risk when `source-path=SCRIPTDIR` is combined with `--external-sources`

## Defense-in-Depth Layers

| Layer | pulse-wrapper.sh (headless) | linters-local.sh (interactive) |
|-------|---------------------------|-------------------------------|
| External sources | No `-x` (never follows sources) | `-x -P SCRIPTDIR` (restricted) |
| RC file | `--norc` (ignores .shellcheckrc) | Uses .shellcheckrc (intentional) |
| Per-file timeout | 30s via timeout/gtimeout | 30s via timeout/gtimeout |
| Memory cap | `ulimit -v 1GB` per subprocess | `ulimit -v 1GB` per subprocess |
| Process guard | `guard_child_processes()` every 60s | N/A (interactive) |

Closes #2875

PR #2884, opened 2026-03-05
t1398.5: Rewrite memory pressure monitor

alex-solovyev:
## Summary

- Rewrites memory pressure monitor to monitor **process-level signals** (individual RSS, runtime, process count) instead of `kern.memorystatus_level` as the primary signal
- Addresses root cause of March 3 kernel panic: aidevops processes (ShellCheck at 5.7 GB, zombie pulses, session accumulation) — not generic OS memory pressure
- Incorporates valid concepts from declined PR #2792 (launchd integration, notifications, cooldown, CLI) with correct thresholds and security fixes

## What Changed

**New file: `.agents/scripts/memory-pressure-monitor.sh`**

| Signal | Threshold | Action |
|--------|-----------|--------|
| Per-process RSS | warn 2 GB, critical 4 GB | Notify per-category with cooldown |
| ShellCheck runtime | > 10 min | Warning notification |
| Other tool runtime | > 30 min | Warning notification |
| Interactive sessions | >= 5 concurrent | Warning notification |
| Aggregate aidevops RSS | >= 8 GB total | Warning notification |
| `kern.memorystatus_level` | Logged only | Secondary/informational — NOT used for alerts |

CLI: `--check` (launchd), `--status`, `--daemon`, `--install`/`--uninstall`, `--help`

**Security fixes vs PR #2792:**
- `_validate_int()` prevents command injection via `$(( ))` arithmetic expansion
- AppleScript `osascript` inputs sanitised (quotes/backslashes escaped)
- All environment variable inputs validated before use
- No XML injection risk in plist generation

**New file: `tests/test-memory-pressure-monitor.sh`** — 28 unit tests covering:
- Configuration validation (valid, non-numeric, below-minimum, leading zeros, injection)
- Duration formatting (seconds, minutes, hours, days, zero)
- Cooldown logic (no file, set+check, clear, invalid content)
- Process age parsing (self, nonexistent PID)
- OS memory info format and values
- CLI commands (help, status, check, unknown)
- Security (env var injection, notification disable)

## Testing

```
Results: 28 passed (total: 28)
```

ShellCheck: zero violations on both files.

Closes #2878

PR #2883, opened 2026-03-05
t1398.4: Add session count awareness

alex-solovyev:
## Summary

- Add `session-count-helper.sh` that detects interactive AI coding sessions (OpenCode, Claude Code, Cursor, Windsurf, Aider) and warns when the count exceeds a configurable threshold (default: 5)
- Distinguishes interactive TUI sessions from headless workers by examining process command lines (`opencode run`, `claude -p` are excluded)
- Integrates into startup flow (`aidevops-update-check.sh`) and pre-edit check (`pre-edit-check.sh`) as non-blocking warnings

## Configuration

- Config key: `safety.max_interactive_sessions` (default: 5, set to 0 to disable)
- Env override: `AIDEVOPS_MAX_SESSIONS`
- Schema (`aidevops-config.schema.json`) and defaults (`aidevops.defaults.jsonc`) updated

## Files Changed

| File | Change |
|------|--------|
| `.agents/scripts/session-count-helper.sh` | **New** — core detection logic with `count`, `check`, `list` commands |
| `.agents/scripts/aidevops-update-check.sh` | Calls session check during startup greeting |
| `.agents/scripts/pre-edit-check.sh` | Non-blocking session warning before edits |
| `.agents/configs/aidevops.defaults.jsonc` | Added `safety.max_interactive_sessions` |
| `.agents/configs/aidevops-config.schema.json` | Added schema for new config key |
| `.agents/scripts/shared-constants.sh` | Added env var mapping for `AIDEVOPS_MAX_SESSIONS` |

## Testing

Verified on Linux with 2 interactive OpenCode sessions + multiple headless workers:

```
$ session-count-helper.sh list
Interactive AI sessions:
  PID 3325002 | OpenCode | 568 MB | uptime: 1-23:50:50
  PID 3742902 | OpenCode | 1149 MB | uptime: 1-20:46:47
Total: 2 interactive | Threshold: 5

$ AIDEVOPS_MAX_SESSIONS=1 session-count-helper.sh check
SESSION_WARNING: 2 interactive AI sessions detected (threshold: 1).
Consider closing unused sessions to reduce memory pressure (~100-400 MB each).
```

Headless workers (`opencode run ...`) are correctly excluded from the count.

Closes #2877

PR #2882, opened 2026-03-05
t1398.3: Add pulse self-watchdog idle detection

alex-solovyev:
## Summary

- Adds process-level idle timeout to `run_pulse()` so the pulse self-terminates when opencode enters idle state, rather than waiting for the full 30-minute stale threshold
- Monitors CPU usage of the entire process tree every 60s — kills when CPU stays below threshold for 5 consecutive minutes (configurable)
- Reduces zombie pulse lifetime from ~30 minutes (wall-clock timeout) to ~7 minutes (2 min startup grace + 5 min idle detection)

## Problem

Root cause: opencode enters idle state after completing the pulse prompt but never exits (sits in a file watcher). The existing `PULSE_STALE_THRESHOLD` (t1397) catches this after 30 minutes, but that's 25 minutes of wasted resources. Worse, if launchd stops firing (sleep, plist unloaded), zombies persist indefinitely — the 28+ hour zombie observed on March 3 was this exact scenario.

## Solution

New `_get_process_tree_cpu()` helper sums `%CPU` for the PID and all descendants. The watchdog loop in `run_pulse()` now tracks consecutive idle seconds alongside the existing wall-clock check:

1. **Wall-clock timeout** (existing, t1397): hard ceiling at `PULSE_STALE_THRESHOLD` (30 min)
2. **Idle detection** (new, t1398.3): kills when CPU < `PULSE_IDLE_CPU_THRESHOLD` (5%) for `PULSE_IDLE_TIMEOUT` (300s) consecutive seconds

Startup grace period of 2 minutes prevents false positives during process initialization.

## New Configuration

| Variable | Default | Description |
|----------|---------|-------------|
| `PULSE_IDLE_TIMEOUT` | `300` (5 min) | Seconds of continuous idle before kill |
| `PULSE_IDLE_CPU_THRESHOLD` | `5` | CPU% below which process is considered idle |

Both are env-overridable and validated through the existing `_validate_int()` security guard.

## Verification

- ShellCheck: zero violations
- Bash syntax check: passes
- All existing functions unchanged — additive change only

Closes #2876

PR #2965, opened 2026-03-06
fix: install memory pressure monitor via setup.sh for reboot survival

marcusquinn:
## Summary

- Memory pressure monitor (`memory-pressure-monitor.sh`) was deployed but not auto-installed by `setup.sh` — it didn't survive reboots or `aidevops update`
- Adds launchd plist generation (macOS, 60s interval, `RunAtLoad=true`) and cron entry (Linux) alongside the existing process guard block
- No consent gate — this is a safety net (auto-kills runaway ShellCheck at >4 GB RSS), same as the process guard

## Context

The t1398 series (PRs #2855, #2881, #2882, #2883, #2884, #2885, #2918, #2930) fixed the root cause of memory blowups (ShellCheck `--external-sources` exponential expansion inside Tabby's bash language server). The memory pressure monitor is the defense-in-depth layer — but it was only installable via `--install` flag, not auto-deployed by `setup.sh`. After a reboot, the monitor wasn't running.

## Verification

- ShellCheck clean on `setup.sh`
- Follows exact same pattern as the process guard block (lines 980-1056)
- Plist matches the one generated by `memory-pressure-monitor.sh --install`
- Monitor confirmed running after manual install: `launchctl list | grep memory-pressure`

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1055,6 +1055,82 @@ GUARD_PLIST
 		fi
 	fi
 
+	# Memory pressure monitor — process-focused memory watchdog (t1398.5, GH#2915).
+	# Monitors individual process RSS, runtime, session count, and aggregate memory.
+	# Auto-kills runaway ShellCheck (language server respawns them). Always installed
+	# when the script exists; no consent needed (safety net, not autonomous action).
+	# macOS: launchd plist (60s interval, RunAtLoad=true) | Linux: cron (every minute)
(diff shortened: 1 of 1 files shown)


marcusquinn:
This PR appears orphaned — no active worker process found and no activity for 6+ hours. Flagging for re-dispatch. If work is still in progress, remove the `status:orphaned` label.

PR #2943, opened 2026-03-06
fix: replace mandatory session gate with layered consent model for pulse

marcusquinn:
## Summary

- Replaces the mandatory session flag gate (`pulse-session.flag`) with a layered consent model that allows the pulse to run via config consent alone, surviving reboots
- Adds stop flag (`pulse-session.stop`) as a pause mechanism that overrides all consent layers
- Updates `pulse-session-helper.sh` status command to show all consent layers and their state

Closes #2942

## Consent Layers (checked in order)

1. **Stop flag** (`~/.aidevops/logs/pulse-session.stop`) — if present, pulse is paused regardless of other layers
2. **Session flag** (`~/.aidevops/logs/pulse-session.flag`) — explicit `aidevops pulse start`, doesn't survive reboots
3. **Config consent** (`orchestration.supervisor_pulse=true` in `config.jsonc`) — persistent, survives reboots

## Files Changed

- `.agents/scripts/pulse-wrapper.sh` — `check_session_gate()` rewritten with layered model
- `.agents/scripts/pulse-session-helper.sh` — `cmd_stop()` creates stop flag, `cmd_status()` shows all layers, `is_config_consent_enabled()` helper added, help text updated

## Testing

- ShellCheck clean on both files
- `aidevops pulse status` shows consent layer breakdown
- `aidevops pulse stop` creates stop flag (overrides config consent)
- `aidevops pulse start` clears stop flag and creates session flag
- With `orchestration.supervisor_pulse=true` and no flags: pulse runs (reboot-safe)

## Summary by CodeRabbit

* **New Features**
  * Enhanced pulse consent management supporting multiple configuration sources with proper precedence handling
  * Added capability to count pulse-enabled repositories

* **Refactor**
  * Improved session gating logic with multi-layer consent verification and comprehensive logging

PR #2966, opened 2026-03-06
hotfix: fix readonly color variable collision in pulse-session-helper.sh

marcusquinn:
## Summary

- Fixes runtime error `readonly variable` when running `aidevops pulse status` (and other pulse-session-helper.sh commands)
- Root cause: `config-helper.sh` (added in PR #2943 follow-up) sources `shared-constants.sh` which declares color variables as `readonly`. The script then tried to redeclare them.
- Fix: use `${VAR+x}` parameter expansion to only define colors when not already set by shared-constants.sh

## Verified

```
$ bash .agents/scripts/pulse-session-helper.sh status
Pulse Status
────────────
  Pulse:       enabled via session flag (aidevops pulse start)
  Consent layers:
    Stop flag:      clear
    Session flag:   active (2026-03-06T01:38:08Z by marcusquinn)
    Config consent: enabled
```


marcusquinn:
@coderabbitai review

PR #2979, opened 2026-03-06
feat: pulse CI failure pattern detection — identify systemic workflow bugs

### LINKED ISSUE #2975: feat: pulse CI failure pattern detection — identify systemic workflow bugs
## Problem

The pulse (Step 3, PRs section) treats all CI failures the same: dispatch a worker to fix it. For per-PR failures (code bugs, test failures), this works. For systemic failures (workflow bugs affecting ALL PRs identically), it's useless — the worker fixes code in the PR, not the workflow, and the same failure recurs on the next PR.

This was the root cause of GH#2973 going undetected: every PR had the same two failing checks (opencode-agent regex false positive + review-bot-gate concurrency cancellation), but the pulse never identified the pattern because it processes PRs individually without cross-PR correlation.

## Proposed Fix

Add a **CI failure pattern detection** step to the pulse, between the current PR processing and issue dispatch. The logic:

1. **Collect failing check names across all open PRs** in the pre-fetched state
2. **Identify checks that fail on 3+ PRs** — these are likely systemic, not per-PR
3. **For systemic failures**: file a single GitHub issue describing the pattern (check name, failure count, sample error), rather than dispatching N workers to fix N PRs individually
4. **For per-PR failures**: continue with the existing dispatch logic

This is a judgment call, not a deterministic rule — the pulse agent should read the check names and failure reasons, correlate across PRs, and decide whether the pattern is systemic. A check that fails on 8/10 PRs with the same error is clearly systemic. A check that fails on 2 PRs with different errors is likely per-PR.

## Implementation Location

Add to `pulse.md` Step 3, as a new subsection "### CI failure pattern detection" between the PR processing and issue dispatch sections. The pulse already has the pre-fetched state with check results per PR — it just needs to correlate them.

## Why This Matters

Without this, the pulse will continue to be blind to workflow-level bugs. The self-improvement signals in AGENTS.md mention "Multiple PRs fail CI with the same error pattern" but the pulse doesn't actually check for this pattern — it's guidance without implementation.

Ref: GH#2973 (the bug this would have caught)

marcusquinn:
## Summary

- Adds a new `### CI failure pattern detection` subsection to `pulse.md` Step 3 that correlates CI failures across PRs to detect workflow-level bugs
- When 3+ PRs share the same failing check name, the pulse files a single workflow-level issue instead of dispatching N individual fix workers
- Modifies the existing "Failing CI" bullet to cross-reference the new detection section before dispatching per-PR fix workers

## Problem

The pulse processed PRs individually without cross-PR correlation. When a workflow bug caused the same CI check to fail on all PRs (e.g., GH#2973: regex false positive + concurrency cancellation), the pulse dispatched individual fix workers for each PR — workers that tried to fix code when the problem was the CI configuration. This wasted worker slots and never fixed the root cause.

## Design decisions

- **Threshold of 3+ PRs**: Guideline, not hard rule — the pulse agent uses judgment (2 PRs with identical errors = probably systemic; 4 PRs with different errors = probably coincidence)
- **On-demand check fetching**: Individual check names are fetched via `gh pr checks` only when 3+ PRs are failing, avoiding unnecessary API calls on normal cycles
- **Rate limited**: Max 2 systemic CI issues per repo per pulse cycle
- **Dedup**: Checks for existing issues before creating new ones
- **Mixed handling**: PRs with both systemic and per-PR failures get fix workers for the per-PR failures only

Closes #2975

## Summary by CodeRabbit

* **Chores**
  * Enhanced CI failure handling with an intelligent pre-dispatch check that distinguishes between systemic workflow issues and per-PR problems, optimizing the dispatch process and reducing unnecessary worker invocations.

### DIFF
--- a/.agents/scripts/commands/pulse.md
+++ b/.agents/scripts/commands/pulse.md
@@ -122,7 +122,7 @@ Then skip to the next PR. The next pulse cycle will retry the permission check �
   - `SKIP` means the PR has a `skip-review-gate` label — it bypasses the bot gate only, NOT the review count requirement.
   - Skip the PR when the formal review count is 0, regardless of bot gate status.
 - **Green CI + zero reviews** → skip this cycle. Zero reviews means "not yet reviewed", NOT "clean to merge". Review bots typically post within 2-5 minutes. The next pulse will pick it up once a review exists.
-- **Failing CI or changes requested** → dispatch a worker to fix it (counts against worker slots)
+- **Failing CI or changes requested** → before dispatching a fix worker, check whether this is a systemic failure (see "CI failure pattern detection" below). If systemic, skip the per-PR dispatch — the workflow-level issue covers it. If per-PR, dispatch a worker to fix it (counts against worker slots).
 
 **For all PRs (regardless of author):**
 


marcusquinn:
This PR appears orphaned — no active worker process found and no activity for 6+ hours. Flagging for re-dispatch. If work is still in progress, remove the `status:orphaned` label.

marcusquinn:
The `Wait for AI Review Bots` check is failing due to a systemic workflow issue — bots are rate-limited and the workflow treats rate-limit notices as non-reviews. Tracked in #3005. This is not a code issue with this PR.

marcusquinn:
Re-ran 'Wait for AI Review Bots' check (stale failure from pre-fix workflow). PR #3006 merged the review-bot-gate rate-limit fix — this re-run uses the updated workflow on main.

marcusquinn:
Re-triggering review-bot-gate after GH#3007 fix (bidirectional prefix matching for CodeRabbit status context).

PR #3123, opened 2026-03-07
GH#3119: Fix command injection in worker sandbox env generation

marcusquinn:
## Summary

- **CRITICAL**: Replace `echo "export VAR='${value}'"` with `printf "export VAR=%q\n" "${value}"` in `generate_sandbox_env()` to prevent command injection via single-quote breakout
- **MEDIUM**: Remove `2>/dev/null` from `rm -rf` in `cleanup_stale_sandboxes()` to improve error observability (`|| true` already prevents script failure)

## Problem

The `generate_sandbox_env` function constructs shell `export` statements by embedding variables (`sandbox_dir`, `REAL_HOME`, `GH_TOKEN`) directly into single-quoted strings. If any variable contains a single quote character (`'`), it terminates the intended string and allows arbitrary command injection when the generated environment lines are subsequently executed. This is a critical vulnerability as it allows arbitrary command execution in the context of the supervisor process, potentially bypassing the sandbox isolation the helper is designed to provide.

## Solution

Use `printf %q` for all variable interpolation in export statements. `printf %q` produces shell-safe escaped output that handles all metacharacters including single quotes, spaces, and special characters. The only `echo` retained is for the literal `AIDEVOPS_SANDBOXED=true` which contains no variable interpolation.

## Verification

- ShellCheck clean (zero violations)
- All 10 `echo "export ..."` lines with variable interpolation replaced with `printf %q`
- The one `echo "export AIDEVOPS_SANDBOXED=true"` (no variable) correctly left as-is

Closes #3119

### DIFF
--- a/.agents/scripts/worker-sandbox-helper.sh
+++ b/.agents/scripts/worker-sandbox-helper.sh
@@ -175,33 +175,36 @@ generate_sandbox_env() {
 	fi
 
 	# Core: override HOME
-	echo "export HOME='${sandbox_dir}'"
+	# Use printf %q to safely escape all shell metacharacters, preventing
+	# command injection if any variable contains single quotes or other
+	# special characters (security fix: GH#3119, PR#3080 review feedback)
+	printf "export HOME=%q\n" "${sandbox_dir}"
 
 	# Preserve REAL_HOME so scripts that need the actual home can find it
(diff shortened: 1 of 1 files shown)


marcusquinn:
@coderabbitai review

Review bots were rate-limited when this PR was created (affected: coderabbitai gemini-code-assist). Requesting a review retry.

PR #3081, opened 2026-03-07
t1412.3: Network domain tiering for worker sandboxing

marcusquinn:
## Summary

- Implements a 5-tier graduated trust model for classifying headless worker network access by domain
- Adds `network-tier-helper.sh` with classify, check, log-access, and report commands
- Integrates with `sandbox-exec-helper.sh` via `--network-tiering` and `--worker-id` flags
- Ships with 97 exact + 45 wildcard domain rules sourced from real session transcript analysis

## Tier Model

| Tier | Action | Examples |
|------|--------|----------|
| 1 | Allow (no log) | `github.com`, `api.github.com` |
| 2 | Allow + log | `registry.npmjs.org`, `pypi.org`, `crates.io` |
| 3 | Allow + log | `sonarcloud.io`, `docs.anthropic.com`, `api.openai.com` |
| 4 | Allow + flag | Any unknown domain (flagged for post-session review) |
| 5 | Deny | `requestbin.com`, `ngrok.io`, raw IPs, `.onion`, `.bit` |

## Files Changed

- **New:** `.agents/configs/network-tiers.conf` — domain classification database (user overrides via `~/.config/aidevops/network-tiers-custom.conf`)
- **New:** `.agents/scripts/network-tier-helper.sh` — core classification engine with self-test, JSONL logging, and reporting
- **Modified:** `.agents/scripts/sandbox-exec-helper.sh` — added `--network-tiering` and `--worker-id` flags with pre-execution domain extraction
- **Modified:** `.agents/tools/security/prompt-injection-defender.md` — added Network Domain Tiering section
- **Modified:** `.agents/tools/ai-assistants/headless-dispatch.md` — added security item #6 and Related links

## Design Decisions

- **File-based lookup** (not bash associative arrays) for macOS bash 3.2 compatibility
- **Best-effort domain extraction** from commands — catches `curl`/`wget` URLs but cannot intercept runtime DNS. Primary value is logging + post-session review, not hard blocking
- **Orthogonal to t1412.1 (fake HOME)** — both can be enabled independently. Network tiering classifies domains; credential isolation restricts what's accessible. Neither depends on the other.
- **User-extensible** — custom config file merges with defaults, last match wins

## Verification

- Built-in self-test (9 cases) passes: `network-tier-helper.sh init`
- ShellCheck clean on both scripts
- Markdown lint clean on both docs
- Tested: classification, logging to access/flagged/denied JSONL, check allow/deny exit codes

Closes #3073

PR #3129, opened 2026-03-07
GH#3114: Fix quality-debt in sandbox-exec-helper.sh from PR #3081 review

marcusquinn:
## Summary

- **HIGH**: Remove `2>/dev/null` from all `network-tier-helper.sh` calls — `|| true` already prevents script exit, so stderr suppression was hiding important errors (missing helper, permission issues, syntax errors)
- **MEDIUM**: Broaden domain extraction from URL-only regex to 3-strategy approach covering bare hostnames (`curl example.com`), git SSH (`git@github.com:user/repo`), and `user@host` patterns (`ssh user@server.example.com`)
- **Bonus**: Fix macOS compatibility bug where `sed 's|https\?://||'` (GNU-only `\?`) silently failed on BSD sed — domains from URLs were never stripped of their protocol prefix

## Changes

**File**: `.agents/scripts/sandbox-exec-helper.sh`

### HIGH: Remove blanket stderr suppression (line ~114-124)

All 4 calls to `$NET_TIER_HELPER` had `2>/dev/null` which hid errors when the helper was missing, not executable, or failed unexpectedly. The `|| true` suffix already prevents `set -e` from aborting the script.

### MEDIUM: Broaden domain extraction (line ~100-122)

Original regex only matched `https?://` URLs. New 3-strategy approach:

| Strategy | Pattern | Example |
|----------|---------|---------|
| 1. URLs | `https?://host` | `curl https://api.github.com/repos` |
| 2. Bare hosts | `curl/wget/nc host.tld` | `wget example.com/file` |
| 3. user@host | `user@host.tld` | `git@github.com:user/repo.git`, `ssh user@server.example.com` |

Includes file extension exclusion filter (`.sh`, `.py`, `.txt`, etc.) to prevent false positives like `scp file.txt` being classified as a domain.

### Bonus: macOS sed compatibility

`sed 's|https\?://||'` → `sed -E 's|https?://||'` — BSD sed (macOS) doesn't support `\?` in basic regex mode.

## Verification

- `bash -n` syntax check: clean
- Tested all 3 extraction strategies against representative commands
- Confirmed no false positives from filenames with dots
- Confirmed edge cases: commands with no domains return empty, flags are skipped

Closes #3114

### DIFF
--- a/.agents/scripts/sandbox-exec-helper.sh
+++ b/.agents/scripts/sandbox-exec-helper.sh
@@ -97,12 +97,31 @@ _sandbox_check_network_tiers() {
 	local command="$1"
 	local wid="$2"
 
-	# Extract potential domains/URLs from the command using common patterns:
-	# - https://domain.com/... or http://domain.com/...
-	# - curl/wget/git clone followed by a URL
-	# - @scope/package from npm (not a domain, skip)
-	local domains
-	domains="$(printf '%s' "$command" | grep -oE 'https?://[a-zA-Z0-9._-]+' | sed 's|https\?://||' | sort -u)" || true
+	# Extract potential domains from the command using multiple strategies:
+	# 1. Full URLs: https://domain.com/... or http://domain.com/...
+	# 2. Bare hostnames after networking tools: curl example.com, wget host.io
+	# 3. Git SSH patterns: git@github.com:user/repo.git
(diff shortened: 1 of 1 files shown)


marcusquinn:
@coderabbitai review

Review bots were rate-limited when this PR was created (affected: coderabbitai gemini-code-assist). Requesting a review retry.

PR #4140, opened 2026-03-11
t1439: fix pulse status drift and worker cap reporting

marcusquinn:
## Summary
- read the wrapper log before the legacy pulse log so pulse status reflects the latest run metadata.
- load config helpers safely in the wrapper and fall back cleanly so max worker reporting stays available.

Closes #4139

## Summary by CodeRabbit

* **Chores**
  * Improved session pulse tracking by checking multiple log sources and reliably reporting "never" when no pulse timestamp is found.
  * Added separate wrapper-level logging to isolate wrapper start events from main logs.
  * Added a defensive fallback for missing configuration so the process continues with a one-time warning instead of failing.

### DIFF
--- a/.agents/scripts/pulse-session-helper.sh
+++ b/.agents/scripts/pulse-session-helper.sh
@@ -21,13 +21,14 @@ export PATH="/bin:/usr/bin:/usr/local/bin:/opt/homebrew/bin:${PATH}"
 
 # Source config-helper for _jsonc_get (shared JSONC config reader)
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-# shellcheck source=config-helper.sh
+# shellcheck source=/dev/null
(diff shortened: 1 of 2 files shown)


marcusquinn:
Dispatching fix worker.
- **Model**: default headless rotation (anthropic/claude-sonnet-4-6 or openai/gpt-5.3-codex)
- **Branch**: bugfix/pulse-status-max-workers
- **Scope**: address changes-requested feedback and stabilize CI
- **Attempt**: 1 of 1
- **Direction**: push fixes to existing branch and keep issue linkage intact

marcusquinn:
Addressed the remaining CodeRabbit feedback in : the fallback  path now emits a one-time warning only when it is actually used, and the message explicitly says default config values are being applied.

marcusquinn:
Addressed the remaining CodeRabbit feedback in `.agents/scripts/pulse-wrapper.sh`: the fallback `config_get` path now emits a one-time warning only when it is actually used, and the message explicitly says default config values are being applied.

PR #4183, opened 2026-03-11
t4154: make email thread reconstruction test grep portable

marcusquinn:
## Summary
- Replace GNU-specific `grep -P` usage in `tests/test-email-thread-reconstruction.sh` with a portable grep pipeline.
- Keep test intent unchanged: fail when markdown links in the generated index contain backslashes.
- Verify with `shellcheck tests/test-email-thread-reconstruction.sh` and `bash tests/test-email-thread-reconstruction.sh`.

Closes #4154

### DIFF
--- a/tests/test-email-thread-reconstruction.sh
+++ b/tests/test-email-thread-reconstruction.sh
@@ -231,7 +231,7 @@ test_relative_paths_cross_directory() {
 	fi
 
 	# Verify paths use forward slashes only (portable Markdown links)
-	if grep '\.md)' "$index_file" | grep -q "\\\\"; then
+	if grep -qE '\(.*\\.*\.md\)' "$index_file"; then
 		print_error "Index links contain backslashes (not portable)"
 		return 1
 	fi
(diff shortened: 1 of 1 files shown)


alex-solovyev:
## Gemini Review Feedback Addressed

Verified the finding and applied the fix in 97e771f6.

**Issue confirmed**: The piped `grep '\.md)' | grep -q "\\\\"` approach was indeed less precise — it would match any line containing both `.md)` and a backslash, even if the backslash was outside the markdown link parentheses (e.g., in a comment or elsewhere on the line).

**Fix applied**: Replaced with a single `grep -qE '\(.*\\.*\.md\)'` that anchors the backslash check *inside* the markdown link parentheses `(...)`, matching the original PCRE intent precisely. `grep -E` (POSIX ERE) is portable across all target systems.

**Verification**: `shellcheck tests/test-email-thread-reconstruction.sh` passes cleanly.

/gemini review

alex-solovyev:
Addressed Gemini review feedback (commit 97e771f6):

- Replaced the two-pipe grep (`grep '\.md)' | grep -q '\\'`) with a single `grep -qE '\(.*\\.*\.md\)'`
- The ERE pattern anchors the backslash check inside the markdown link parentheses, eliminating false positives from backslashes elsewhere on the line
- Pattern matches Gemini's suggestion exactly and is POSIX-portable

Re-review requested.

PR #4532, opened 2026-03-14
t1480: fix pulse sessions overlapping — run wrapper from repo path (GH#4513)

### LINKED ISSUE #4513: fix: pulse sessions overlapping — launchd fires new pulse before prior exits
## Observation

During pulse at 2026-03-14T00:08Z, 6 concurrent pulse sessions were detected running simultaneously (PIDs: 3095551, 3245841, 3373539, 3381328, 3443412, 3507937 + the current session).

## Root Cause Hypothesis

The launchd scheduler fires a new pulse every 2 minutes regardless of whether the prior pulse has exited. When a pulse takes >2 minutes (e.g., due to slow `gh` API calls or worker dispatch), the next pulse starts with a stale pre-fetched state snapshot, potentially seeing dead workers as alive and skipping valid dispatch candidates.

## Impact

1. Stale state: pre-fetched snapshot showed 24 worker PIDs (from a prior cycle) — most were dead by the time this pulse ran
2. Wasted resources: 5+ pulse sessions consuming context/API quota simultaneously
3. Risk of double-dispatch: if circuit breaker misses a race window, two pulses could dispatch the same issue

## Proposed Fix

Add a pulse lock file (`~/.aidevops/logs/pulse.lock`) with PID. On startup, check if the lock PID is still alive — if yes, exit immediately (another pulse is running). Release lock on exit. This is a standard single-instance guard pattern.

Alternatively, use `launchd` `StartInterval` with a longer interval (5 min) and rely on the lock for safety.

## Evidence

```
ps aux | grep 'sandbox-exec-helper.sh run' | grep pulse
# Shows 6 concurrent pulse processes
```

alex-solovyev:
## Summary

- **Root cause**: The cron entry ran the deployed `~/.aidevops/agents/scripts/pulse-wrapper.sh` instead of the repo version. Fixes merged to main (like the flock guard from GH#4409/t1478) were not effective until the next `setup.sh` run, creating a deployment gap that allowed 6 concurrent pulse sessions.
- **Fix**: Change `wrapper_script` in `setup.sh` to point to `${_pulse_repo_dir}/.agents/scripts/pulse-wrapper.sh` (the repo version). Since the canonical repo directory is always on `main` per the worktree-first workflow, this ensures fixes are immediately effective after merge.
- **After merge**: Run `./setup.sh` once to update the cron entry to use the new repo path. Future fixes will be effective immediately without needing `setup.sh`.

## Changes

| File | Change |
|------|--------|
| `setup.sh` | Changed `wrapper_script` from deployed path to repo path; updated guard comment |
| `.agents/scripts/pulse-wrapper.sh` | Updated header comment to reflect new invocation path |

## Verification

- `shellcheck setup.sh` — zero violations
- `shellcheck .agents/scripts/pulse-wrapper.sh` — zero violations
- The cron entry will now reference the repo version, ensuring the flock guard (GH#4409) and all future fixes are immediately effective

Closes #4513

## Summary by CodeRabbit

* **Chores**
  * Updated pulse wrapper configuration to execute directly from the repository, enabling immediate deployment of fixes after merge without requiring setup.sh re-execution.
  * Updated documentation clarifying that the pulse wrapper is invoked by cron (Linux) or launchd (macOS) every 120 seconds, now running from the repo source directly.

### DIFF
--- a/.agents/scripts/pulse-wrapper.sh
+++ b/.agents/scripts/pulse-wrapper.sh
@@ -39,7 +39,10 @@
 #   where launchd fires between rm -f and the next write, which caused the
 #   82-concurrent-pulse incident (2026-03-13T02:06:01Z, issue #4318).
 #
-# Called by launchd every 120s via the supervisor-pulse plist.
+# Called by cron (Linux) or launchd (macOS) every 120s.
+# GH#4513: The cron/launchd entry now runs this script directly from the
+# repo (not the deployed copy at ~/.aidevops/agents/scripts/), so fixes
+# are effective immediately after merge without needing setup.sh.
 
 set -euo pipefail
 
--- a/setup.sh
+++ b/setup.sh
@@ -857,11 +857,19 @@ main() {
 	#   - Consent persisted: written to config.jsonc so it survives updates
 	#   - Never silently re-enabled: if config says false, skip entirely
 	#   - Non-interactive: only installs if config explicitly says true
-	local wrapper_script="$HOME/.aidevops/agents/scripts/pulse-wrapper.sh"
+	#
 	local pulse_label="com.aidevops.aidevops-supervisor-pulse"
 	local _aidevops_dir _pulse_repo_dir
 	_aidevops_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 	_pulse_repo_dir=$(_resolve_main_worktree_dir "$_aidevops_dir")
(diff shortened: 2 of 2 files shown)


marcusquinn:
Closing: superseded by PR #4525 which was already merged. PR #4525 addressed GH#4513 with the flock→mkdir-based atomic lock approach. This PR's approach (running wrapper from repo path) is a different fix for the same issue — if still needed, it should be filed as a separate follow-up issue.

PR #4540, opened 2026-03-14
fix(pulse-wrapper): add mkdir atomic lock fallback for flock-unavailable systems (GH#4513)

### LINKED ISSUE #4513: fix: pulse sessions overlapping — launchd fires new pulse before prior exits

alex-solovyev:
## Summary

- Replaces the weak `flock` fallback (silent fall-through to racy PID dedup) with a two-layer lock strategy
- Layer 1: `flock(1)` on FD 9 — primary, atomic, crash-safe, auto-releases on SIGKILL
- Layer 2: `mkdir` atomic lock — fallback for systems without `flock` (macOS, containers), also active as defence-in-depth when `flock` is available

## Root Cause (GH#4513)

The `flock` guard added in GH#4409 had a silent fallback: when `flock(1)` was unavailable, it logged a warning and returned 0 (success), falling through to PID-based dedup. PID dedup has a TOCTOU race window — multiple instances can read the PID file simultaneously before any writes its own PID.

The incident at 2026-03-14T00:08Z (6 concurrent pulse sessions) had two contributing factors:
1. The deployed `~/.aidevops/agents/scripts/pulse-wrapper.sh` predated the GH#4409 flock fix (`setup.sh` not re-run after merge), so all instances used the racy PID fallback
2. Even with the flock fix deployed, the fallback path was unsafe for systems without `flock`

## Fix

`mkdir` is atomic on all POSIX-compliant local filesystems — the kernel guarantees that only one process succeeds when multiple race to create the same directory. The lock dir (`pulse-wrapper.lock.d/`) contains a PID file; stale locks (dead owner PID) are cleaned up on startup. A `trap` removes the lock dir on normal exit.

Decision table:
| flock available | mkdir succeeds | Action |
|---|---|---|
| yes | yes | Proceed (both layers held) |
| yes | no, owner alive | Proceed under flock authority (log WARN) |
| yes | no, owner dead | Clean stale lock, retry mkdir, proceed |
| no | yes | Proceed (mkdir layer only) |
| no | no, owner alive | Exit — genuine concurrent instance |
| no | no, owner dead | Clean stale lock, retry mkdir, proceed |

## Testing

- ShellCheck passes with zero warnings
- Logic verified against all 6 decision table cases above

Closes #4513

### DIFF
--- a/.agents/scripts/pulse-wrapper.sh
+++ b/.agents/scripts/pulse-wrapper.sh
@@ -5,7 +5,9 @@
 # but never exits, blocking all future pulses via the pgrep dedup guard.
 #
 # This wrapper:
-#   1. flock-based instance lock prevents concurrent pulses (GH#4409)
+#   1. Two-layer instance lock prevents concurrent pulses (GH#4409, GH#4513):
+#      - flock(1) on FD 9 (primary, Linux/util-linux — atomic, crash-safe)
+#      - mkdir atomic lock (fallback, all POSIX — closes flock-unavailable gap)
 #   2. Uses a PID file with staleness check (not pgrep) for dedup
 #   3. Cleans up orphaned opencode processes before each pulse
 #   4. Kills runaway processes exceeding RSS or runtime limits (t1398.1)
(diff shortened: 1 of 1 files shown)


marcusquinn:
Closing: GH#4513 was already resolved by PR #4525 (merged). That PR replaced flock with mkdir-based atomic lock. This PR's two-layer approach may have merit as a follow-up improvement, but the original issue is closed. If the two-layer strategy is desired, please file a new issue.

PR #4634, opened 2026-03-14
fix(t3287): address PR #2154 review feedback in setup-modules/core.sh

alex-solovyev:
## Summary

Addresses all 3 unactioned review findings from PR #2154 flagged in issue #3287.

### Fixes applied

**HIGH (gemini) — sudo credential caching before spinner**
Added `sudo -v` before the `case` statement for package managers that require root (`apt`, `dnf`, `yum`, `pacman`, `apk`). `run_with_spinner` backgrounds the command with output redirected to `/dev/null`, so a `sudo` password prompt would be hidden and the script would appear to hang indefinitely. Caching credentials first prevents this.

**HIGH (coderabbit) — propagate update failures with warning**
`brew update` and `apt-get update` failures are now caught with `if !` and a `print_warning` message is shown. The install still proceeds (the user may have a usable local index), but they get a clear diagnostic instead of a confusing downstream install error.

**MEDIUM (gemini) — explicit `return 0` at end of `install_packages()`**
Added `return 0` after the `case` block to satisfy the repository style guide requiring explicit return statements in all functions.

### Verification

- `shellcheck setup-modules/core.sh` — zero violations
- All three findings from the issue body are addressed

Closes #3287

### DIFF
--- a/setup-modules/core.sh
+++ b/setup-modules/core.sh
@@ -219,16 +219,24 @@ install_packages() {
 	shift
 	local packages=("$@")
 
+	# Cache sudo credentials before backgrounding any sudo commands.
+	# run_with_spinner redirects output to /dev/null, so a sudo password prompt
+	# would be hidden and the script would appear to hang indefinitely.
+	[[ "$pkg_manager" =~ ^(apt|dnf|yum|pacman|apk)$ ]] && sudo -v
+
 	case "$pkg_manager" in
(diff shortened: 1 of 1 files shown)


marcusquinn:
Closing as duplicate — PR #4632 targets the same issue (t3287) and was opened first. Merging #4632 instead.

PR #5220, opened 2026-03-18
t5215: revert grep portability change — grep -qiE is already POSIX-compliant

marcusquinn:
## Summary

- Reverts the `tr | grep` pipeline introduced in the original commit back to `grep -qiE`.
- `grep -qiE` is POSIX-compliant (`-i` and `-E` are both specified in POSIX.1-2008); the original issue #5215 finding was incorrect.
- The `tr | grep` pipeline was less readable, introduced BRE escaped-pipe complexity (`\|`), and added unnecessary process overhead.
- Addresses Gemini's inline review comment on line 182 of PR #5220.

## Verification
- `shellcheck .agents/scripts/tests/test-tier3-simplified.sh` — zero violations
- All 9 `tr '[:upper:]' '[:lower:]' | grep -q` occurrences reverted to `grep -qiE`

Closes #5215

### DIFF
--- a/.agents/scripts/tests/test-tier3-simplified.sh
+++ b/.agents/scripts/tests/test-tier3-simplified.sh
@@ -179,7 +179,7 @@ test_full_loop_helper() {
 	# Test: start with no prompt — must fail AND produce an error message
 	local rc=0
 	output=$("$helper" start "" 2>&1) || rc=$?
-	if [[ $rc -ne 0 ]] && echo "$output" | grep -qiE "no prompt|usage|required|missing|empty"; then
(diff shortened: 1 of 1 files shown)


marcusquinn:
Gemini review feedback: the `tr | grep` pipeline introduced by this PR is less readable than the original `grep -qiE`. Gemini confirms `grep -qiE` is POSIX-compliant. The fix should revert to the original pattern or use Gemini's suggestion directly. Dispatching a fix worker.

marcusquinn:
Closing: the fix worker correctly determined that grep -qiE is already POSIX-compliant. The original issue #5215 finding was incorrect — no code change needed. Net diff vs main is empty.

PR #5341, opened 2026-03-20
fix: address CodeRabbit and Gemini review findings in oauth-pool.mjs

marcusquinn:
## Summary

Follow-up to #5337 — addresses the CHANGES_REQUESTED review findings that arrived after the original PR was merged.

## Changes

- **`registerPoolProvider` migration robustness** (CodeRabbit, line 1456): The previous `!config.provider[id]` check only handled the absent-provider case. If a provider existed from a previous version but had no models (empty `models` object), the check would pass and the provider would remain broken (invisible in the auth dialog). Fixed to also update providers that exist but have no models.

- **Default case error message** (CodeRabbit, line 1664): The `Unknown action` message listed available actions but omitted `assign-pending`. Added to the list.

- **`assignPendingToken` property updates** (Gemini medium, line 551): Replaced 6 individual property assignments with `Object.assign` for cleaner, more idiomatic code.

- **Token data deduplication** (Gemini medium, lines 1107/1367): Extracted `tokenData` const in both `createPoolAuthHook` and `createOpenAIPoolAuthHook` to eliminate repeated `Date.now() + expires_in * 1000` expressions across `upsertAccount`, `savePendingToken`, `client.auth.set`, and the return value.

## Testing

No behaviour change — these are correctness and code quality fixes. The migration fix is the only functional change: it ensures users upgrading from a version that stored no models will have their pool providers correctly updated on next startup.

### DIFF
--- a/.agents/plugins/opencode-aidevops/oauth-pool.mjs
+++ b/.agents/plugins/opencode-aidevops/oauth-pool.mjs
@@ -540,15 +540,18 @@ function assignPendingToken(provider, email) {
   if (idx < 0) return false;
 
   // Update the existing account with the pending token
-  pool[provider][idx].refresh = pending.refresh;
-  pool[provider][idx].access = pending.access;
-  pool[provider][idx].expires = pending.expires;
-  pool[provider][idx].lastUsed = new Date().toISOString();
-  pool[provider][idx].status = "active";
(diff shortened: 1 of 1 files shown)


marcusquinn:
Merge conflict after #5342 merged (both touch oauth-pool.mjs). Dispatching worker to rebase.

PR #5384, opened 2026-03-21
feat: add Cursor support to oauth-pool-helper.sh and /models-pool-check

marcusquinn:
## Summary

- Add `oauth-pool-helper.sh add cursor` — reads credentials from the local Cursor IDE installation (SQLite state DB or auth.json). No browser flow needed — instant, just requires being logged into Cursor IDE.
- Update `/models-pool-check` command doc with Cursor-specific guidance: different setup flow (instant, no browser), expired tokens are normal for Cursor (short-lived, auto-refreshed by plugin)
- Cross-platform support: macOS, Linux, Windows paths for Cursor state DB

## Testing

- `shellcheck` clean
- Tested on macOS: successfully read credentials from Cursor state.vscdb, decoded JWT for email/expiry, saved to pool.json
- `check` command shows Cursor account alongside Anthropic accounts
- `list` and `help` updated with Cursor references

PR #5401, opened 2026-03-21
GH#5380: simplify open_browser() with if/elif/else chain

alex-solovyev:
## Summary

- Refactors `open_browser()` in `oauth-pool-helper.sh` to combine the command-existence check and execution into a single `if/elif/else` condition chain
- If a browser opener exists but fails to launch the URL, the function now falls through to try the next alternative instead of silently swallowing the error with `|| true`
- Addresses PR #5377 review feedback from Gemini Code Assist

Closes #5380

### DIFF
--- a/.agents/scripts/oauth-pool-helper.sh
+++ b/.agents/scripts/oauth-pool-helper.sh
@@ -106,12 +106,12 @@ save_pool() {
 # Open URL in browser (best-effort, never fatal)
 open_browser() {
 	local url="$1"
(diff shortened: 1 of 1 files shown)


marcusquinn:
Closing this PR due to merge conflict with PR #5384 (feat: add Cursor support to oauth-pool-helper.sh), which touches the same file and was created earlier. The simplification in this PR should be re-applied after #5384 merges. Re-opening issue #5380 for dispatch after #5384 is merged.

PR #5412, opened 2026-03-21
t5391: Fix quality-debt review feedback on provider-auth.mjs

alex-solovyev:
## Summary

Addresses CodeRabbit review feedback from PR #5359 on `provider-auth.mjs`:

- **Harden unknown status sorting**: Changed `?? 2` to `?? 99` so any future unknown account status (e.g., `disabled`) is assigned the lowest priority and tried last, rather than defaulting to `rate-limited` priority
- **Remove dead code**: Deleted unused `TOKEN_ENDPOINT` and `CLIENT_ID` constants that duplicated the canonical definitions in `oauth-pool.mjs` and were never referenced in this file

Closes #5391


marcusquinn (approved):
All CI checks pass. Quality-debt fix for provider-auth.mjs — hardening unknown status sorting with ?? 99 is correct.

PR #5437, opened 2026-03-21
fix: remove escaped $PORT in app-store-connect.md localdev example

marcusquinn:
## Summary

- Removes the backslash escape from `\$PORT` on line 210 of `.agents/tools/mobile/app-store-connect.md`
- The escaped form caused the shell to treat `$PORT` as a literal string rather than substituting the environment variable, which would cause `localdev-helper.sh` to pass the literal text `$PORT` as the port argument instead of the actual port number
- Consistent with the same command pattern in `services/hosting/local-hosting.md`

Closes #5430

**Source**: Gemini review finding on PR #5407 (discussion_r2969932979)


alex-solovyev:
@coderabbitai review

PR #5441, opened 2026-03-21
t5432: fix(oauth-pool): consolidate python3 calls in cmd_add_cursor

marcusquinn:
## Summary

Addresses the two medium-severity findings from gemini-code-assist on PR #5384.

- **Finding 1** (`oauth-pool-helper.sh:397`): The auth.json token extraction used a single `python3` call but then spawned two subshells (`head -1` / `tail -1`) to split the output. Replaced with a bash 3.2-compatible `while IFS= read -r` loop that captures both tokens in one pass — no extra subshells, no extra file reads.

- **Finding 2** (`oauth-pool-helper.sh:447`): The JWT decode used two separate `python3` invocations — first to decode the JWT payload into a JSON string, then to parse that JSON string. Merged into a single call that outputs email and exp directly, eliminating the intermediate JSON serialisation round-trip and second interpreter startup.

Both fixes use `while IFS= read -r` loops instead of `mapfile` (which is bash 4.0+ and forbidden on macOS's default bash 3.2).

ShellCheck passes with zero violations.

Closes #5432

### DIFF
--- a/.agents/scripts/oauth-pool-helper.sh
+++ b/.agents/scripts/oauth-pool-helper.sh
@@ -386,16 +386,28 @@ cmd_add_cursor() {
 	# Source 1: cursor-agent auth.json
 	if [[ -f "$cursor_auth_json" ]]; then
 		print_info "Reading from Cursor auth.json..."
-		local auth_tokens
-		auth_tokens=$(python3 -c "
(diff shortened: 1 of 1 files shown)


alex-solovyev:
@coderabbitai review

PR #5444, opened 2026-03-21
fix: re-enable auth hook as single object — restores Ctrl+A add account

marcusquinn:
## Summary

Re-enables the Anthropic pool auth hook by using a single object instead of an array. OpenCode v1.2.27 supports `auth: singleHookObject` but crashes on `auth: [array]`.

The auth hook was disabled entirely in v3.1.35 because we thought all auth hooks were broken. The actual bug was only with the array format — single objects always worked.

**Restores**: Ctrl+A "Add Account" flow for Anthropic Pool in the TUI.

**Still via shell**: OpenAI and Cursor accounts are added via `oauth-pool-helper.sh add openai|cursor`.

Closes the regression from v3.1.35.


alex-solovyev:
@coderabbitai review

PR #5525, opened 2026-03-23
fix: add mid-session 429 pool rotation in provider-auth.mjs

marcusquinn:
## Summary

- **Bug:** `provider-auth.mjs` custom fetch wrapper passed HTTP 429 (rate limit) responses through unhandled. `fetchWithPoolFailover()` existed in `oauth-pool.mjs` but was exported and never called by anything.
- **Fix:** Intercepts 429 responses mid-session, marks the current account as rate-limited, rotates to the next available pool account, and retries the request once with new credentials.
- **Impact:** Users with 2+ OAuth pool accounts now get automatic mid-session rotation when one account hits rate limits, instead of the session failing.

## What was done

- Added `injectPoolToken` import from `oauth-pool.mjs`
- Added `RATE_LIMIT_COOLDOWN_MS` constant (60 seconds)
- After the `fetch()` call, added 429 interception logic that:
  1. Identifies the current account by matching the access token
  2. Marks it `rate-limited` with a 1-minute cooldown in `oauth-pool.json`
  3. Finds the next available pool account (active, not in cooldown, LRU order)
  4. Refreshes the token via `ensureValidToken()` if needed
  5. Injects the new token into the built-in provider via `injectPoolToken()`
  6. Retries the request once with the new account's credentials
  7. If no alternate account is available, logs and returns the original 429

## How it was tested

- Syntax validation: `node --check provider-auth.mjs` passes
- Logic review: follows the same pattern as the existing (but unused) `fetchWithPoolFailover()` in `oauth-pool.mjs`
- Pool state verified: 2 active Anthropic accounts in `oauth-pool.json` with valid tokens

## Files changed

- `.agents/plugins/opencode-aidevops/provider-auth.mjs` — added 429 interception and pool rotation logic (+78 lines)

## Key decisions

- Used 1-minute cooldown (matching `RATE_LIMIT_COOLDOWN_MS` in `oauth-pool.mjs`) rather than parsing `Retry-After` headers — keeps the logic simple and the cooldown short enough for interactive use
- Retry only once per 429 — prevents infinite rotation loops if all accounts are rate-limited
- Reused existing `injectPoolToken()` and `ensureValidToken()` rather than duplicating logic

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resiliency against temporary rate limits: the system now automatically rotates to alternate accounts and retries requests when a rate limit is encountered, reducing failed requests and interruptions.
  * Added cooldown handling for rate-limited accounts and better diagnostics so recovery and failover are more reliable and observable.

### DIFF
--- a/.agents/plugins/opencode-aidevops/provider-auth.mjs
+++ b/.agents/plugins/opencode-aidevops/provider-auth.mjs
@@ -9,12 +9,17 @@
  *   - User-Agent matching current Claude CLI version
  *   - Deprecated beta header filtering
  *   - Pool token injection on session start
+ *   - Mid-session 429 rotation: on rate limit, marks current account as
+ *     rate-limited, rotates to next pool account, and retries once
  *
  * The pool (oauth-pool.mjs) manages multiple account tokens.
  * This module makes the built-in provider use them correctly.
  */
 
-import { ensureValidToken, getAccounts, patchAccount, getAnthropicUserAgent } from "./oauth-pool.mjs";
+import { ensureValidToken, getAccounts, patchAccount, getAnthropicUserAgent, injectPoolToken } from "./oauth-pool.mjs";
+
+/** Default cooldown when rate limited mid-session (ms) — 1 minute */
+const RATE_LIMIT_COOLDOWN_MS = 60_000;
 
 const TOOL_PREFIX = "mcp_";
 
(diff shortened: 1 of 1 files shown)


marcusquinn:
Dispatching fix worker for CodeRabbit critical findings.
- **Model**: sonnet (anthropic/claude-sonnet-4-6)
- **Branch**: bugfix/pool-rotation-429-mid-session
- **Scope**: Fix token mismatch (injectPoolToken using wrong alt) and add try/catch around ensureValidToken to prevent exception bypassing 429 return
- **Attempt**: 1 of 1
- **Direction**: Apply CodeRabbit's proposed diff from PR #5525 inline comment

alex-solovyev:
## Verification of CodeRabbit fixes (commit 9687fc1)

**Critical finding — Token mismatch + unhandled exception**: Fixed.

1. `ensureValidToken()` wrapped in try/catch (lines 307-312) — on exception, logs error and continues to next alternate account. If all alternates fail, the original 429 response is returned to the caller.

2. `injectPoolToken(client, currentEmail)` replaced with direct `client.auth.set()` using the validated `alt` account's tokens (lines 320-328). This ensures the **same** account is used for both provider auth storage and the request Authorization header. The `client.auth.set()` API shape matches exactly what `injectPoolToken()` uses internally (verified against `oauth-pool.mjs:1271-1279`), but without the LRU re-selection that caused the mismatch.

3. `client.auth.set()` also wrapped in try/catch (lines 329-332) — defensive addition so provider injection failures fall through to the next alternate.

4. `injectPoolToken` import removed from line 19.

**Nitpick — `patchAccount` race condition**: Acknowledged as pre-existing. The non-atomic read-modify-write in `patchAccount` predates this PR and affects all pool operations, not just 429 rotation. Not fixing in this PR scope — filed as a future hardening consideration.

All CI checks pass. Ready for merge.

PR #5572, opened 2026-03-23
fix: check all non-bot users in auto-decline safety net, not just owner

alex-solovyev:
## Summary

- **Fixes the high-severity Gemini review finding from PR #5522**: the safety-net auto-decline `jq` filter in `draft-response-helper.sh` was restricting the comment search to only `$username` (the authenticated user), but the feature's goal is to block auto-decline when *any* non-bot user has commented
- Removed the `select(.user.login == "${username}")` filter and kept only the bot-exclusion filter (`select(.user.login | test("\\[bot\\]$"; "i") | not)`), so comments from any human user prevent auto-decline
- Without this fix, a comment from a different non-bot user on the notification issue would be missed, leading to incorrect auto-decline of drafts that have active human engagement

Closes #5559

### DIFF
--- a/.agents/scripts/draft-response-helper.sh
+++ b/.agents/scripts/draft-response-helper.sh
@@ -832,15 +832,19 @@ cmd_check_approvals() {
 			fi
 		fi
 
-		# Check for user comments on this notification issue
+		# Check for comments from ANY non-bot user on this notification issue.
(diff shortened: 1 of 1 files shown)


alex-solovyev:
Closing as duplicate — issue #5559 already has PR #5576 open. Flagging the duplicate per pulse protocol.

PR #5612, opened 2026-03-23
fix: handle 401/403 (server-side token revocation) in OAuth pool with auto-refresh

marcusquinn:
## Summary

- **Root cause**: Anthropic revokes OAuth access tokens server-side before the local expiry timestamp (tokens claim 8h validity but are rejected after ~1-2h idle). Three users independently hit `invalid x-api-key` on session start after a break.
- **provider-auth.mjs**: Added 401/403 handler — force-refreshes the current account's token first, then rotates to next pool account if refresh fails (same retry-once pattern as existing 429 handler)
- **oauth-pool.mjs `ensureValidToken()`**: Added proactive refresh when token has used >50% of its claimed lifetime, preventing the gap between local expiry and server-side revocation. Graceful fallback: if proactive refresh fails, the existing token is still returned

## What was broken

1. `ensureValidToken()` only checked `account.expires > Date.now()` — if the local timestamp said "valid", it returned the stale token without testing it against the API
2. `provider-auth.mjs` fetch wrapper handled 429 (rate limit) with rotation, but did NOT handle 401/403 (invalid token) — the error propagated straight to the user
3. `fetchWithPoolFailover()` in oauth-pool.mjs DID handle 401/403, but was never called from provider-auth.mjs

## Testing

- Verified both refresh tokens still work against Anthropic's token endpoint
- Deployed to `~/.aidevops/agents/plugins/` for immediate testing
- Next session start will exercise the proactive refresh path

## Summary by CodeRabbit

* **New Features**
  * Proactive token refresh to refresh access tokens before local expiry
  * Merged auth hook to ensure provider requests use proper Bearer authorization

* **Bug Fixes**
  * Mid-session recovery with a single automatic retry and fallback account rotation on auth errors
  * 5-minute cooldown applied to accounts after authentication failures
  * Safer handling when proactive refreshes fail (preserves existing token instead of forcing error)

### DIFF
--- a/.agents/plugins/opencode-aidevops/index.mjs
+++ b/.agents/plugins/opencode-aidevops/index.mjs
@@ -1165,11 +1165,26 @@ export async function AidevopsPlugin({ directory, client }) {
     // Phase 6: LLM observability — capture assistant message metadata (t1308)
     event: async (input) => handleEvent(input),
 
-    // Phase 7: OAuth multi-account pool (t1543, t1548, t1549)
-    // OpenCode v1.2.27 only supports auth as a single object, not an array.
-    // Using the Anthropic pool hook as the primary — it's the most-used provider.
-    // OpenAI and Cursor accounts are added via oauth-pool-helper.sh instead.
-    auth: createPoolAuthHook(client),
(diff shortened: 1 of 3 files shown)


alex-solovyev:
@coderabbitai review

PR #5678, opened 2026-03-24
GH#5677: Fix memory-helper.sh recall FTS5 query failure

alex-solovyev:
## Summary

- Fix `.param set` double-quote expansion in heredoc that caused all `memory-helper.sh recall` queries to fail with `fts5: syntax error near ""`
- Convert shared (global DB) search from direct string interpolation to parameterized `.param set` binding for consistency and SQL injection safety

## Root Cause

In `recall.sh` line 262, `.param set :query "$escaped_query"` used double quotes inside a heredoc (`<<EOF`). When `$escaped_query` contained FTS5-quoted tokens like `"bottled"`, the heredoc expansion consumed the double quotes — sqlite3 received `.param set :query bottled` (no quotes) or an empty string, causing the FTS5 syntax error.

## Fix

Changed to single quotes: `.param set :query '${escaped_query}'`. Single quotes inside a heredoc are literal characters (not shell quoting), so sqlite3 receives the full FTS5 query string with its double-quoted tokens intact.

## Verification

Tested with sqlite3 FTS5:
- Single-word query `"bottled"` (was failing with syntax error) — now returns correct result
- Multi-word query `"custom" "water"` — returns correct result
- Old double-quote pattern confirmed to reproduce the exact error from the issue

ShellCheck: zero violations.

Closes #5677

## Summary by CodeRabbit

* **Bug Fixes**
  * Made memory recall/search more robust: query handling and escaping improved so shared/global memory lookups return fields and filtered results consistent with main memory queries.

* **Chores**
  * Incremented internal task counter.

PR #5684, opened 2026-03-24
GH#5677: Preserve apostrophes in FTS5 .param set quoting

### LINKED ISSUE #5677: Bug: memory-helper.sh recall fails for all queries due to .param set double-quote expansion in heredoc
## Summary

`memory-helper.sh recall --query "custom water"` returns `Runtime error near line 2: fts5: syntax error near ""` for all queries (single-word and multi-word).

## Root Cause

In `scripts/memory/recall.sh` around line 263, the FTS5 query is bound via:

```bash
.param set :query "$escaped_query"
```

When `$escaped_query` contains double-quoted FTS5 tokens like `"custom" "water"`, the heredoc expansion produces:

```
.param set :query "custom" "water"
```

sqlite3 CLI interprets this as: set `:query` to `custom` (the first double-quoted token), then `"water"` is an extra argument that gets ignored. The bound value is just `custom` (without quotes), which is valid FTS5. But for single-word queries like `"bottled"`, the expansion produces `.param set :query "bottled"` which sqlite3 sees as setting `:query` to an empty string (the double quotes are consumed as shell quoting), causing the FTS5 syntax error.

## Fix

Change the `.param set` line to use single quotes:

```bash
.param set :query '${escaped_query}'
```

Single quotes prevent the shell from interpreting the double quotes inside `$escaped_query`, so sqlite3 receives the full FTS5 query string correctly.

## Verified Fix

```bash
escaped_query='"custom" "water"'
sqlite3 ':memory:' <<EOF
CREATE VIRTUAL TABLE t USING fts5(content);
INSERT INTO t VALUES ('custom water bottled');
.param set :query '${escaped_query}'
SELECT content FROM t WHERE t MATCH :query;
EOF
# Output: custom water bottled  ✓
```

## Impact

All `memory-helper.sh recall` calls fail. The `--recent` mode works (no FTS5 query). Memory store works. Only recall/search is broken.

## Discovered in

t009 (mainframe) — cross-session memory seeding session. Memories stored successfully but recall verification failed.

## File

`~/.aidevops/agents/scripts/memory/recall.sh` line ~263

alex-solovyev:
## Summary

- Replace lossy apostrophe-stripping workaround with proper SQL literal escaping for sqlite3's `.param set` dot-command
- Queries like `O'Reilly` now preserve the apostrophe instead of being split into `O Reilly`
- Uses SQL single-quoted literal wrapped in double-quoted dot-command argument per sqlite3 CLI docs

## Problem

PR #5678 merged with a workaround that stripped apostrophes from queries before tokenization (`O'Reilly` → `O Reilly`). While functionally correct for FTS5 matching (the tokenizer splits on apostrophes anyway), this was lossy and didn't address the root cause identified by CodeRabbit and Gemini reviewers.

**Root cause**: sqlite3's dot-command parser does NOT support `''` escaping inside single-quoted arguments. The dot-command parser is separate from the SQL parser. So `.param set :query '"O''Reilly"'` breaks at the first embedded `'`.

## Fix

Build the `.param set` value as an SQL single-quoted literal, then escape it for the dot-command's double-quoted argument context:

```
.param set :query "'\"O''Reilly\" \"guide\"'"
```

- Outer `"..."`: dot-command argument boundary (supports `\"` escaping)
- Inner `'...'`: SQL text literal delimiters
- `''`: SQL-escaped apostrophe (evaluated by SQL parser, not dot-command parser)
- `\"`: dot-command-escaped double quote (preserves FTS5 token quoting)

Uses `sed` for the escaping pipeline to avoid bash quoting conflicts between parameter expansion and heredoc contexts.

## Verification

All 10 test cases pass:
- Simple multi-word, single word, hyphenated tokens
- **O'Reilly**, **can't**, **it's O'Reilly's book** (apostrophe cases)
- Backslash stripping, JSON output mode
- ShellCheck: zero violations

## Review Comments Addressed

Addresses all 5 unresolved review comments from PR #5678:
1. **CodeRabbit (line 262)**: `.param set` quoting — now uses SQL literal in double-quoted dot-command arg
2. **CodeRabbit (line 332)**: shared-mode filter propagation — already fixed in #5678, preserved
3. **Gemini critical (line 262)**: FTS5 syntax correctness — preserved and improved
4. **Gemini security-high (line 319)**: parameterized binding for shared query — preserved
5. **Gemini security-high (line 332)**: `:query` parameter vs interpolation — preserved

Closes #5677

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced search functionality to correctly process queries containing special characters such as apostrophes and backslashes. Previously, these characters could cause unexpected behavior; now they are properly handled throughout the search pipeline, ensuring users receive accurate results regardless of special characters in their input.

### DIFF
--- a/.agents/scripts/memory/recall.sh
+++ b/.agents/scripts/memory/recall.sh
@@ -192,14 +192,10 @@ cmd_recall() {
 	# Quote each token individually to handle special chars (hyphens = NOT in FTS5)
 	# "foo-bar baz" → "\"foo-bar\" \"baz\"" (each token quoted, joined by implicit AND)
 	#
-	# Apostrophes and backslashes are replaced with spaces before tokenization.
-	# Apostrophes inside FTS5 double-quoted tokens cannot be SQL-escaped (doubling
-	# them to '' produces invalid FTS5 syntax). Backslashes are not valid FTS5 query
-	# syntax and cause a parse error. Replacing both with spaces splits e.g.
-	# "O'Reilly" into "O" and "Reilly", which still matches stored content because
-	# FTS5 tokenizes stored content the same way (GH#5678).
-	local query_clean="${query//"'"/" "}"
-	query_clean="${query_clean//\\/  }"
+	# Backslashes are stripped before tokenization — they are not valid FTS5 query
+	# syntax and cause a parse error. Apostrophes are preserved through proper
+	# quoting (see param_query construction below).
+	local query_clean="${query//\\/  }"
 	local tokenised_query=""
 	local token
 	set -f # Disable globbing during tokenization (SC2086)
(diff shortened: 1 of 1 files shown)


johnwaldo:
Pulse: All CI checks passed, CodeRabbit APPROVED, review-bot gate PASS. Ready for maintainer merge.

PR #6303, opened 2026-03-25
GH#5861: Guard against missing option values in instantly-helper.sh

alex-solovyev:
## Summary

- Adds `require_option_arg()` helper that checks `$#` >= 2 before each `shift 2` in the argument parsing loop
- Prevents `shift 2` from aborting the script under `set -euo pipefail` when an option value is missing (e.g., `--campaign-id` with no value)
- Provides a clear error message ("Option --campaign-id requires a value") instead of a cryptic shell abort

Addresses the inline review comment on PR #6284.

Closes #5861

### DIFF
--- a/.agents/scripts/instantly-helper.sh
+++ b/.agents/scripts/instantly-helper.sh
@@ -361,6 +361,18 @@ cmd_request() {
 	return $?
 }
 
+require_option_arg() {
+	local option="$1"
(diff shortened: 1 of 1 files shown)


marcusquinn:
Superseded by #6284 which already merged the same refactoring for instantly-helper.sh. Closing as duplicate.

PR #6414, opened 2026-03-25
GH#6413: Fix sandbox-exec-helper.sh timeout enforcement with secondary watchdog

alex-solovyev:
## Summary

- Add secondary watchdog to `_sandbox_exec_with_pgkill` that independently tracks wall-clock time via `date(1)` and kills the child process group if the deadline is exceeded — defense-in-depth against parent crash, OOM, or sleep drift
- Add process state filtering in `list_active_worker_processes` to exclude zombie (`Z`) and stopped (`T`) processes from active worker counts, preventing inflated struggle ratios and stale worker detection
- Update test fixtures and add new test for zombie/stopped process filtering (all 8 tests pass)

## Details

**Root cause**: When the sandbox wrapper process crashes or gets OOM-killed, the child process (running in its own session via `setsid`) survives indefinitely. The primary polling loop dies with the parent, so no timeout enforcement occurs. Observed: PID 88026 ran 18+ hours despite `--timeout 3600`.

**Secondary watchdog design**:
- Spawned as a background subshell alongside the primary polling loop
- Sleeps in 30s chunks (responsive to parent kill signals)
- Uses `date(1)` wall-clock verification, not just sleep duration
- 10% grace period (min 5s, max 60s) avoids racing with the primary loop
- Killed on normal exit via `_pgkill_cleanup` to prevent orphaned watchers

**Process state filtering**:
- `ps axo pid,stat,etime,command` now includes the `stat` column
- awk filters exclude processes starting with `Z` (zombie) or `T` (stopped)
- Output format preserved as `pid etime command` for backward compatibility
- `SN` (sleeping, low priority) processes are NOT excluded — they are valid running states

Closes #6413

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved system process monitoring by filtering out zombie and stopped processes for accurate counts
  * Enhanced sandbox timeout enforcement with a dedicated watchdog mechanism

* **Tests**
  * Extended test coverage for zombie and stopped process filtering scenarios

### DIFF
--- a/.agents/scripts/pulse-wrapper.sh
+++ b/.agents/scripts/pulse-wrapper.sh
@@ -1353,14 +1353,28 @@ list_active_worker_processes() {
 	#      (sandbox-disabled path and test fixtures)
 	# Exclude: lines starting with "node " (node child) or whose command
 	# starts with a path ending in "/.opencode " (binary grandchild).
-	ps axo pid,etime,command | awk '
+	#
+	# GH#6413: Process state filtering — exclude zombie (Z) and stopped (T)
+	# processes. These are dead/stuck processes that hold no useful work but
+	# appear as "active" in worker counts, inflating struggle ratios and
+	# preventing the pulse from dispatching replacements. The stat column is
+	# used for filtering only; output format remains pid,etime,command for
+	# backward compatibility with all consumers.
+	ps axo pid,stat,etime,command | awk '
 		/\/full-loop/ &&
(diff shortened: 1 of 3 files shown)


marcusquinn (approved):
All CI checks green. CodeRabbit findings verified as addressed: exit code 124 normalization implemented via watchdog marker file (lines 543-548), PID recycling guard addressed in commit 274b41d. Merging.

PR #6677, opened 2026-03-26
fix: add TZ=UTC to macOS date conversion in backoff_active_for_key

### LINKED ISSUE #6549: Bug: headless-runtime-helper.sh backoff expiry check fails due to timezone offset in date conversion
## Observed

In `backoff_active_for_key()`, the retry_after timestamp (stored as UTC ISO 8601, e.g. `2026-03-26T03:41:12Z`) is converted to epoch using:

```bash
retry_epoch=$(date -j -f '%Y-%m-%dT%H:%M:%SZ' "$stored_retry_after" '+%s')
```

On macOS, `date -j -f` interprets the input as **local time**, not UTC. This causes the epoch to be offset by the local timezone (e.g., UTC-6 adds 21600 seconds), making an expired backoff appear to still be active.

## Impact

All configured models appear backed off even after the retry window has passed. Workers cannot be dispatched. Pulse silently fails to launch workers with exit code 75 (EX_TEMPFAIL). Observed during pulse 2026-03-26T03:42Z — backoff expired at 03:41Z UTC but was still active due to this bug.

## Root cause

`date -j -f '%Y-%m-%dT%H:%M:%SZ'` on macOS does not honor the trailing `Z` (UTC indicator). The format string `%Z` would capture the timezone but `%S` followed by literal `Z` just strips the Z without applying UTC offset.

## Fix

Use `TZ=UTC date -j -f '%Y-%m-%dT%H:%M:%SZ'` to force UTC interpretation on macOS, or use `date -u -d` on Linux. The existing fallback `date -u -d` (Linux) is correct — only the macOS path is broken.

johnwaldo:
## Summary

- Fixes a timezone offset bug in `backoff_active_for_key()` where `date -j -f '%Y-%m-%dT%H:%M:%SZ'` on macOS interprets the UTC ISO 8601 timestamp as local time, inflating the epoch by the local UTC offset (e.g. UTC-5 adds 18000 seconds)
- An expired backoff appeared still active, causing all workers to exit with EX_TEMPFAIL (75) and the pulse to silently fail to dispatch
- Fix: prefix the macOS `date -j` call with `TZ=UTC` so the timestamp is correctly parsed as UTC; the Linux `date -u -d` fallback was already correct and is unchanged
- Adds a regression test that writes an already-expired UTC `retry_after` directly into the DB and verifies the model is selectable again (25/25 tests pass)

## Root cause

`date -j -f '%Y-%m-%dT%H:%M:%SZ'` on macOS does not honour the trailing `Z` as a UTC indicator — the format string strips the literal `Z` but does not apply a UTC offset. The fix uses `TZ=UTC date -j -f ...` to force UTC interpretation.

## Testing

- All 25 existing tests pass
- New regression test `Backoff Expiry UTC Timezone (GH#6549)` passes
- Manually verified: without fix, a UTC-5 system inflates epoch by 18000s (backoff appears active); with fix, epoch is correct (backoff correctly expired)

## Files changed

- `.agents/scripts/headless-runtime-helper.sh` — one-character fix: add `TZ=UTC ` prefix to macOS date call (line 458)
- `tests/test-headless-runtime-helper.sh` — regression test for the UTC timezone expiry check

Closes #6549

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed UTC timezone handling in retry backoff logic to ensure consistent and accurate retry timing calculations.

* **Tests**
  * Added regression test for backoff expiry UTC timezone handling.

### DIFF
--- a/.agents/scripts/headless-runtime-helper.sh
+++ b/.agents/scripts/headless-runtime-helper.sh
@@ -455,7 +455,7 @@ backoff_active_for_key() {
 	if [[ -n "$stored_retry_after" ]]; then
 		local now_epoch retry_epoch
 		now_epoch=$(date -u '+%s')
-		retry_epoch=$(date -j -f '%Y-%m-%dT%H:%M:%SZ' "$stored_retry_after" '+%s' 2>/dev/null || date -u -d "$stored_retry_after" '+%s' 2>/dev/null || printf '%s' "0")
+		retry_epoch=$(TZ=UTC date -j -f '%Y-%m-%dT%H:%M:%SZ' "$stored_retry_after" '+%s' 2>/dev/null || date -u -d "$stored_retry_after" '+%s' 2>/dev/null || printf '%s' "0")
 		if [[ "$retry_epoch" -le "$now_epoch" ]]; then
 			clear_provider_backoff "$key"
 			return 1
--- a/tests/test-headless-runtime-helper.sh
+++ b/tests/test-headless-runtime-helper.sh
@@ -297,6 +297,39 @@ fi
 # Clean up
 bash "$HELPER" backoff clear anthropic >/dev/null 2>&1 || true
 
+section "Backoff Expiry UTC Timezone (GH#6549)"
+# Regression test: backoff_active_for_key() must treat stored retry_after as UTC.
+# On macOS, date -j -f without TZ=UTC interprets the timestamp as local time,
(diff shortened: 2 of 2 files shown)


johnwaldo:
Closing as superseded by PR #6683 which addresses the same issue (#6549) with the same fix. PR #6683 is from the maintainer and will be merged once CI completes.

PR #6688, opened 2026-03-26
fix: guard all unbound variables in full-loop-helper.sh background/status paths

marcusquinn:
## Summary

Fixes three classes of `unbound variable` crashes in `full-loop-helper.sh` under `set -u`, as reported in #6684.

### 1. `STARTED_AT` missing from `load_state` allowlist (line 267 crash)

`load_state()` parses YAML frontmatter via an allowlist `case` statement. `STARTED_AT` was not in the allowlist, so it was never populated from the state file. `cmd_status` and `cmd_resume` then referenced `$STARTED_AT` on an unbound variable.

**Fix:** Added `STARTED_AT` to the allowlist in `load_state`.

### 2. Optional flag variables unbound on `export` (background path crash)

Variables like `MAX_TASK_ITERATIONS`, `SKIP_PREFLIGHT`, etc. were only assigned when the corresponding CLI flag was passed. The background path unconditionally `export`s all of them — under `set -u`, exporting an unset variable crashes.

**Fix:** Pre-initialise all optional flag variables at script level with safe defaults (matching the existing `DEFAULT_*` constants).

### 3. `pid_file` local variable referenced in EXIT trap (line 356 crash)

`_run_foreground` declared `local pid_file=...` then used it in `trap '... "$pid_file"' EXIT`. In bash, `local` variables are scoped to the function — when the EXIT trap fires after the function returns, `$pid_file` is unbound.

**Fix:** Replaced with a script-level variable `_RUN_FG_PID_FILE` that survives the function's scope.

### 4. State variables pre-initialised at script level

`CURRENT_PHASE`, `STARTED_AT`, `PR_NUMBER`, `SAVED_PROMPT` are now initialised to empty strings at the top of the script, so `cmd_status`/`cmd_resume` are safe even if `load_state` is skipped or returns early.

## Testing

- `shellcheck` passes (zero new violations; SC1091 is pre-existing info-level)
- Simulated the exact crash scenario from the issue: created a state file and ran `status` — output is correct, exit 0
- Script loads cleanly under `set -euo pipefail` with no flags set

Closes #6684

### DIFF
--- a/.agents/scripts/full-loop-helper.sh
+++ b/.agents/scripts/full-loop-helper.sh
@@ -14,7 +14,25 @@ readonly STATE_FILE="${STATE_DIR}/full-loop.local.state"
 readonly DEFAULT_MAX_TASK_ITERATIONS=50 DEFAULT_MAX_PREFLIGHT_ITERATIONS=5 DEFAULT_MAX_PR_ITERATIONS=20
 readonly BOLD='\033[1m'
 
+# Initialise all optional state variables with safe defaults so set -u never
+# fires on them, regardless of which flags were passed or whether load_state
+# was called.
 HEADLESS="${FULL_LOOP_HEADLESS:-false}"
+MAX_TASK_ITERATIONS="${MAX_TASK_ITERATIONS:-$DEFAULT_MAX_TASK_ITERATIONS}"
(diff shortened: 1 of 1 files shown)


marcusquinn:
Closing — issue #6684 was already resolved by PR #6687 which was merged to main while this PR was in CI. The fixes are equivalent; #6687 landed first.

PR #6810, opened 2026-03-27
GH#6704: Fix zsh IFS tab compatibility in agent-generated inline commands

### LINKED ISSUE #6704: Bug: zsh IFS=$'\t' on while-read breaks $() command resolution — affects all tab-parsing loops
## Summary

When a script runs in zsh (macOS default, used by the MCP Bash tool) and uses `IFS=$'\t'` on a `while read` line to parse tab-separated output, **all `$()` command substitutions inside the loop body fail with "command not found"** — even for standard utilities like `sed`, `grep`, `cut`, `basename`.

This is a zsh-specific behavior. Bash scopes `IFS` on the `read` line to the `read` command only. Zsh leaks it into the loop body's subshells, corrupting PATH-based command resolution.

## Reproduction

```zsh
# FAILS in zsh — sed not found
echo -e "100\ttest.md" | while IFS=$'\t' read -r size path; do
  base=$(echo "$path" | sed 's|\.md$||')
  echo "base=$base"
done
# Output: zsh:3: command not found: sed

# WORKS in bash
/bin/bash -c '
echo -e "100\ttest.md" | while IFS=$'"'"'\t'"'"' read -r size path; do
  base=$(echo "$path" | sed "s|\.md$||")
  echo "base=$base"
done
'
# Output: base=test
```

## Impact on aidevops

This pattern is common in aidevops scripts that parse `gh api` or `jq` tab-separated output:

```bash
gh api '...' --jq '"\(.size)\t\(.path)"' | while IFS=$'\t' read -r size path; do
  base=$(echo "$path" | sed 's|.*/||')  # FAILS in zsh
  if echo "$CACHE" | grep -qi "$base"; then  # FAILS in zsh
    continue
  fi
done
```

**Observed in this session:** The autonomous simplification pulse script's dedup logic silently failed — `sed` and `grep` calls inside the while-read loop returned "command not found", so the dedup check never filtered anything. This caused the pulse to re-dispatch the same file (CHAPTER-12.md) 4+ times per cycle, creating duplicate worktrees and wasting tokens. Filed separately as GH#6696.

**Scope:** Any aidevops helper script that (a) runs through the MCP Bash tool (which uses the user's default shell, typically zsh on macOS), and (b) uses `IFS=$'\t' read` to parse structured output, is affected. Scripts with `#!/bin/bash` shebangs are safe when run directly, but NOT when their logic is inlined into MCP Bash tool calls.

## Fix

**For inline Bash tool commands (agent-generated):** Use zsh-safe alternatives inside while-read loops:

```zsh
# SAFE: parameter expansion (no subshell)
while IFS=$'\t' read -r size path; do
  base="${path##*/}"      # instead of $(basename "$path")
  base="${base%.md}"      # instead of $(echo "$path" | sed 's|\.md$||')
done

# SAFE: avoid IFS=$'\t' entirely — use cut in a subshell-free way
while read -r line; do
  size="${line%%$'\t'*}"
  path="${line#*$'\t'}"
done
```

**For framework scripts (`.sh` files):** Already safe if they have `#!/bin/bash` or `#!/usr/bin/env bash` shebangs and are invoked directly. No change needed.

**For build.txt / agent guidance:** Add a note under "Bash 3.2 Compatibility" or a new "zsh Compatibility" section warning agents not to use `IFS=$'\t'` with `$()` command substitutions when generating inline shell commands. Recommend parameter expansion (`${var##*/}`, `${var%.ext}`) over `sed`/`basename` inside loops.

## Classification

This is a **framework guidance issue**, not a code bug. The framework scripts themselves are fine (they use bash shebangs). The problem is that AI agents generate inline shell commands for the MCP Bash tool, which runs in zsh. The agents need to know about this zsh pitfall to avoid it.

alex-solovyev:
## Summary

- Add zsh compatibility guidance to `build.txt` documenting the `IFS=$'\t'` scoping difference between bash and zsh that breaks `$()` command substitutions inside while-read loops
- Provide safe alternatives: parameter expansion, IFS save/restore, tab-free parsing patterns
- Recommend helper script invocation over inline tab-parsing logic for the MCP Bash tool

Closes #6704

## Root Cause

zsh (macOS default shell, used by the MCP Bash tool) leaks `IFS=$'\t'` from `while ... read` lines into the loop body's subshells. This corrupts PATH-based command resolution — all `$()` substitutions fail with "command not found" for standard utilities (sed, grep, cut, basename). Bash scopes IFS to the `read` command only, so this is a zsh-specific behavioral difference.

## Scope

- **74 instances** of `IFS=$'\t' read` exist across framework `.sh` scripts — all have `#!/usr/bin/env bash` shebangs and are **unaffected** when invoked directly
- The problem only manifests in **agent-generated inline commands** that run through the MCP Bash tool (which uses the user's default shell, typically zsh on macOS)
- This is a **framework guidance issue**, not a code bug — agents need to know about this pitfall

## Changes

| File | Change |
|------|--------|
| `.agents/prompts/build.txt` | New "zsh Compatibility" section (38 lines) after existing "Bash 3.2 Compatibility" section |

## Runtime Testing

- **Testing level**: `self-assessed`
- **Risk classification**: Low — documentation/guidance only, no code logic changes
- **Rationale**: This adds agent prompt guidance. No executable code is modified. The 74 existing script instances are confirmed safe (all have bash shebangs).

## Summary by CodeRabbit

* **Documentation**
  * Added guidance on shell compatibility considerations for users working with the MCP Bash tool, including recommendations for handling inline commands in alternative shell environments and best practices for shell script invocation.

### DIFF
--- a/.agents/prompts/build.txt
+++ b/.agents/prompts/build.txt
@@ -466,12 +466,14 @@ Git is the primary audit trail for all work. Every change should be discoverable
 #
 # Critical difference: IFS=$'\t' on while-read loops
 # In bash: `IFS=$'\t'` on the `while` line is scoped to the `read` command only.
-# In zsh: `IFS=$'\t'` LEAKS into the loop body's subshells, corrupting PATH-based
-# command resolution. ALL $() command substitutions inside the loop fail with
-# "command not found" — even for standard utilities (sed, grep, cut, basename).
+# In zsh: `IFS=$'\t'` LEAKS into the loop body's subshells, causing $() command
+# substitutions and external commands to fail with "command not found" — even for
+# standard utilities (sed, grep, cut, basename). The exact mechanism may vary by
+# zsh version, but the observed behaviour is consistent: commands that work
+# outside the loop fail inside it when IFS is set to tab on the while-read line.
 #
 # BROKEN in zsh (agent-generated inline commands):
-#   echo -e "100\ttest.md" | while IFS=$'\t' read -r size path; do
+#   printf '100\ttest.md\n' | while IFS=$'\t' read -r size path; do
 #     base=$(echo "$path" | sed 's|\.md$||')   # FAILS: sed not found
 #     if echo "$x" | grep -qi "$base"; then     # FAILS: grep not found
 #       continue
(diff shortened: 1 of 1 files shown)


alex-solovyev:
Superseded by PR #6817 which addresses the review feedback. Closing this PR.

PR #6788, opened 2026-03-27
fix: remove library deps from PIP_TOOLS in tool-version-check.sh

johnwaldo:
## Summary

- Removes `crawl4ai` and `dspy-ai` from `PIP_TOOLS` in `tool-version-check.sh` (Option A, maintainer-approved)
- These are project-level library dependencies managed by `pyproject.toml` / `requirements.txt` inside project venvs — not global CLI tools
- Adds an explanatory comment documenting the exclusion rationale and linking to the issue
- Updates two inline comments that referenced the now-removed libraries

## Problem

The auto-updater was installing library packages into `~/.local/` (global pip scope), creating redundant copies that diverge from pinned project venv versions. In one reported case this accumulated 99 redundant packages. There was also an internal contradiction: `requirements.txt` pins dspy with explicit security comments ("Update versions deliberately, not automatically"), yet `tool-version-check.sh` was auto-updating it globally.

## Changes

- `PIP_TOOLS` now contains only CLI tools: `beads-viewer`, `analytics-mcp`, `outscraper-mcp-server`
- Comment block after `PIP_TOOLS` explains why library deps are excluded
- Updated `get_pip_installed_version` and `check_tool` comments to remove stale references to the removed packages

## Runtime Testing

- **Risk classification**: Low — docs/config change to a shell script array definition
- **Testing level**: self-assessed
- **Verification**: ShellCheck passes with zero violations

Closes #6763

## Summary by CodeRabbit

* **Chores**
  * Improved the development tool management system by refining how dependencies are organized and tracked at different levels. Certain Python packages are now properly excluded from global version checks and updates, as they should be managed within the project's local environment. Updated related documentation comments to clarify this improved dependency management approach.

PR #6814, opened 2026-03-27
fix: remove library deps from PIP_TOOLS in tool-version-check.sh

marcusquinn:
## Summary

- Removes `dspy-ai` and `crawl4ai` from `PIP_TOOLS` in `tool-version-check.sh` (Option A, maintainer-approved in #6763)
- These are project-level library dependencies managed inside project venvs via `pyproject.toml` / `requirements.txt` — not global CLI tools
- Adds an explanatory comment documenting the exclusion rationale and linking to the issue, with accurate claims (no false assertions about which manifest files contain these packages)

## Problem

The auto-updater was installing library packages into `~/.local/` (global pip scope), creating redundant copies that diverge from pinned project venv versions. One user accumulated 99 redundant packages. There was also an internal contradiction: `requirements.txt` pins `dspy` with explicit security comments ("Update versions deliberately, not automatically"), yet `tool-version-check.sh` was auto-updating it globally.

A prior PR (#6788 from a fork) addressed this but had a CodeRabbit finding: the comment incorrectly claimed `crawl4ai` was present in `pyproject.toml`/`requirements.txt` when it is not. This PR implements the same fix with an accurate comment.

## Changes

- `PIP_TOOLS` now contains only CLI tools: `beads-viewer`, `analytics-mcp`, `outscraper-mcp-server`
- Comment after `PIP_TOOLS` explains why library deps are excluded, without making false claims about which manifest files contain them

## Runtime Testing

- **Risk classification**: Low — shell array definition change, no logic modified
- **Testing level**: self-assessed
- **Verification**: ShellCheck passes with zero violations

Closes #6763

### DIFF
--- a/.agents/scripts/tool-version-check.sh
+++ b/.agents/scripts/tool-version-check.sh
@@ -132,11 +132,14 @@ BREW_TOOLS=(
 
 PIP_TOOLS=(
 	"pip|Beads Viewer|beads_viewer|--version|beads-viewer|$(_pip_upgrade_cmd beads-viewer)"
-	"pip|DSPy|dspy|--version|dspy-ai|$(_pip_upgrade_cmd dspy-ai)"
-	"pip|Crawl4AI|crawl4ai|--version|crawl4ai|$(_pip_upgrade_cmd crawl4ai)"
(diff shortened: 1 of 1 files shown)


alex-solovyev:
PR is CONFLICTING with main — needs rebase. Will dispatch a fix worker when a slot opens.

PR #17505, opened 2026-04-06
GH#17442: fix systemd pulse TimeoutStartSec — use PULSE_STALE_THRESHOLD_SECONDS

### LINKED ISSUE #17442: bug: systemd oneshot pulse has no TimeoutStartSec — hung wrapper blocks all future timer activations
## Description

The systemd service unit generated by `_install_pulse_systemd()` uses `Type=oneshot` but does not set `TimeoutStartSec`. Per systemd.service(5), oneshot services default this to **infinity**: "except when Type=oneshot is used, in which case the timeout is disabled by default."

If the pulse wrapper hangs (stuck `gh api` call, network stall, deadlocked child process), systemd waits forever. The `.timer` unit queues subsequent activations that cannot run until the hung invocation completes — resulting in a total pulse outage from a single hung process.

**Affected lines** (`setup-modules/schedulers.sh`):
- Lines 457–466: service unit `[Service]` section — no `TimeoutStartSec` directive

## Root Cause

The original Linux support commit `99025ef18` (t1748) chose `Type=oneshot` correctly (pulse is fire-and-forget), but the infinite timeout default for oneshot services is non-obvious and was not addressed. No subsequent fix has added it.

**Why the wrapper self-watchdog doesn't help**: The wrapper's `PULSE_STALE_THRESHOLD` watchdog (pulse-wrapper.sh line 604) kills *child* processes it spawned. If the hang occurs in the main wrapper process before the watchdog loop starts (e.g., during `gh auth status` check, credential resolution, or lock acquisition), the watchdog never runs.

## Expected Behavior

The service should have a hard timeout ceiling so systemd SIGTERM/SIGKILLs the hung process, unblocking the timer for the next cycle. This matches launchd behavior where overlapping `StartInterval` invocations are simply skipped — a hung process never blocks future invocations.

## Suggested Fix

Add `TimeoutStartSec=900` (or match `PULSE_STALE_THRESHOLD`, currently 1800 in the launchd plist / 900 wrapper default) to the `[Service]` section:

```ini
[Service]
Type=oneshot
TimeoutStartSec=900
ExecStart=/bin/bash ${wrapper_script}
```

This is distinct from GH#17438 (`KillMode=control-group` killing background workers on exit) — that bug is about *how* processes are killed on normal exit; this bug is about the service never exiting at all.

## Environment

- **aidevops version**: 3.6.100
- **OS**: Ubuntu 24.04.4 LTS (audit target: systemd 255)
- **Shell**: bash 5.2.21(1)-release
- **gh CLI**: 2.89.0

---
[aidevops.sh](https://aidevops.sh) v3.6.100 plugin for [OpenCode](https://opencode.ai) v1.3.15 with claude-opus-4-6

marcusquinn:
## Summary

- The systemd service unit for the supervisor pulse was using `TimeoutStartSec=120` (the 2-minute timer interval), which would SIGTERM a legitimately running pulse worker mid-dispatch
- A pulse worker can run for up to `PULSE_STALE_THRESHOLD_SECONDS` (1800s) before the wrapper's own watchdog kills it; the systemd timeout must be at least as large
- Fix: pass `PULSE_STALE_THRESHOLD_SECONDS` as `timeout_sec` so the generated `.service` file contains `TimeoutStartSec=1800`, matching the wrapper's stale threshold

## Root Cause

`setup_supervisor_pulse()` called `_install_scheduler_linux` with `timeout_sec="120"` (arg 13). This propagated through `_install_scheduler_systemd` to `TimeoutStartSec=120` in the generated unit file. The `_install_scheduler_systemd` function already supports `TimeoutStartSec` correctly — the bug was the wrong value being passed at the call site.

## Files Changed

- `setup-modules/schedulers.sh`: line 232 — `"120"` → `"${PULSE_STALE_THRESHOLD_SECONDS}"`

## Runtime Testing

**Risk level**: Low — single-line change to a shell variable substitution in a scheduler setup function. No logic change; only the timeout value written to the generated `.service` file changes. The fix is self-assessed: the variable `PULSE_STALE_THRESHOLD_SECONDS=1800` is defined at line 10 of the same file and already passed as an env var to the service, so the substitution is guaranteed to resolve correctly.

## Key Decisions

- Used `${PULSE_STALE_THRESHOLD_SECONDS}` (the existing constant) rather than hardcoding `900` or `1800`, keeping the timeout in sync with the constant definition at the top of the file
- Did not change `interval_sec` (arg 5 = `"120"`) — the 2-minute timer interval is correct and unrelated to the service timeout

Closes #17442

---
[aidevops.sh](https://aidevops.sh) v3.6.110 plugin for [OpenCode](https://opencode.ai) v1.3.15 with claude-sonnet-4-6

## Summary by CodeRabbit

* **Chores**
  * Scheduler installation process now dynamically uses a configurable stale threshold parameter instead of a hardcoded fixed value. This enhancement provides improved flexibility during system deployment and configuration, allowing administrators to adjust staleness detection mechanisms for different operational environments while maintaining consistent monitoring behavior across installations.

### DIFF
--- a/setup-modules/schedulers.sh
+++ b/setup-modules/schedulers.sh
@@ -229,7 +229,7 @@ PULSE_STALE_THRESHOLD=${PULSE_STALE_THRESHOLD_SECONDS}"
 				"true" \
 				"false" \
 				"" \
-				"120"
+				"${PULSE_STALE_THRESHOLD_SECONDS}"
 		fi
 	elif [[ "$_pulse_lower" == "false" && "$_pulse_installed" == "true" ]]; then
 		# User explicitly disabled but pulse is still installed — clean up


marcusquinn:
**What**: Fixed `TimeoutStartSec` in the systemd supervisor pulse service unit — changed from `120` (2-min interval) to `PULSE_STALE_THRESHOLD_SECONDS` (1800s).

**Issue**: Closes #17442

**Files changed**: `setup-modules/schedulers.sh` (1 line)

**Testing**: ShellCheck passes. Self-assessed low-risk: single variable substitution change, no logic change.

**Key decisions**: Used the existing `PULSE_STALE_THRESHOLD_SECONDS` constant rather than hardcoding a value, keeping timeout in sync with the stale threshold definition.

---
Merged via PR #17505 to main.
_Merged by deterministic merge pass (pulse-wrapper.sh)._
---
[aidevops.sh](https://aidevops.sh) v3.6.110 spent 6m on this as a headless bash routine.

PR #17685, opened 2026-04-07
fix: avoid stat output pollution and WAL stdout leak

saschabuehrle:
Fixes #17683

On Linux, the macOS stat path was writing errors to stdout before the fallback, which could break size arithmetic under strict shell settings. file_size_bytes() now picks the stat variant by platform.

Also moved PRAGMA journal_mode=WAL out of the schema heredoc and run it with stdout suppressed, so wal doesn't leak into command output.

Greetings, saschabuehrle

## Summary by CodeRabbit

* **Chores**
  * Improved cross-platform compatibility for file size detection on macOS and Linux systems.
  * Enhanced database write-ahead logging initialization process.

### DIFF
--- a/.agents/scripts/opencode-db-archive.sh
+++ b/.agents/scripts/opencode-db-archive.sh
@@ -144,7 +144,11 @@ file_size_bytes() {
 		echo "0"
 		return 0
 	fi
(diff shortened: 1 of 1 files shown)


alex-solovyev (approved):
Auto-approved by pulse — collaborator PR (author: @saschabuehrle). All pre-merge checks passed.

Issue #18056, opened 2026-04-10
CI nesting threshold proximity: 245/247 violations (2 headroom)


alex-solovyev:
## CI Nesting Threshold Proximity Warning

The shell nesting depth violation count is within **5** of the CI threshold.

- **Current violations**: 245
- **CI threshold**: 247 (from `.agents/configs/complexity-thresholds.conf`)
- **Headroom remaining**: 2

### Why this matters

The `Complexity Analysis` CI check fails when nesting depth violations exceed the threshold. When PRs add new scripts with deep nesting, they push the count over the threshold and block all open PRs. This happened 6 times in a short window (GH#17808).

### Recommended actions

1. Reduce nesting depth in the highest-depth scripts (run `complexity-scan-helper.sh scan` to identify them)
2. Or bump the threshold in `.agents/configs/complexity-thresholds.conf` with a documented rationale

### Files to check

Run locally to see current violators:
```bash
git ls-files '*.sh' | while read -r f; do
  d=$(awk 'BEGIN{d=0;m=0} /^[[:space:]]*#/{next} /[[:space:]]*(if|for|while|until|case)[[:space:]]/{d++;if(d>m)m=d} /[[:space:]]*(fi|done|esac)[[:space:]]*$||/^[[:space:]]*(fi|done|esac)$/{if(d>0)d--} END{print m}' "$f")
  [ "$d" -gt 8 ] && echo "$d $f"
done | sort -rn | head -20
```

---
[aidevops.sh](https://aidevops.sh) v3.6.227 plugin for [OpenCode](https://opencode.ai) spent 13m on this as a headless bash routine.

marcusquinn:
**Stale assignment recovered** (GH#15060)

Previously assigned to: alex-solovyev
Reason: no dispatch claim comment found, no recent activity
Threshold: 600s

The assigned runner had no active worker process and produced no progress within the threshold. Unassigned and relabeled `status:available` for re-dispatch.

_This recovery prevents the orphaned-assignment deadlock where offline runners permanently block all dispatch._

marcusquinn:
DISPATCH_CLAIM nonce=b22c2eab13ab57ef8552944a2c7cf852 runner=marcusquinn ts=2026-04-10T00:48:35Z max_age_s=1800

marcusquinn:
Dispatching worker (deterministic).
- **Worker PID**: 64519
- **Model**: openai/gpt-5.4
- **Tier**: standard
[...]

PR #18046, opened 2026-04-10
feat: optimise headless worker dispatch — pre-create worktrees, skip unnecessary calls

marcusquinn:
## Summary

Optimise headless worker dispatch to reduce token waste and improve task completion rates, particularly for non-Anthropic models (gpt-5.4) that are more sensitive to context overhead.

**Problem:** Analysis of 12 gpt-5.4 worker sessions (issues #18031-#18042) showed workers spending 40-60% of their token budget on framework ceremony before touching any code. Only 1 of 12 reached the commit stage. The #1 blocker was worktree creation (5-8 tool calls), followed by unnecessary update checks and session-rename calls.

**Changes:**

- **Pre-create worktree in dispatcher** (`pulse-wrapper.sh`): `dispatch_with_dedup()` now creates the worktree before launching the worker and passes `WORKER_WORKTREE_PATH`, `WORKER_WORKTREE_BRANCH`, `WORKER_ISSUE_NUMBER` as env vars. Workers start in the worktree ready to code. Falls back gracefully if creation fails.
- **Export `HEADLESS=1`** to worker environment so scripts can detect headless mode.
- **Skip update check** (`aidevops-update-check.sh`): Early-exit when `HEADLESS=1` — workers never need version checks or security advisories. Saves ~24K tokens of output per session.
- **Fast-path pre-edit-check** (`pre-edit-check.sh`): When `WORKER_WORKTREE_PATH` is set and matches current directory, immediately return exit 0. Eliminates the entire worktree detection/creation flow.
- **Deterministic session title**: Use `issue_title` instead of generic `"Issue #NNN"` for searchable session history without workers calling `session-rename`.
- **HEADLESS_CONTINUATION_CONTRACT V6**: Updated to tell workers their worktree is pre-created, skip update-check/session-rename/worktree-helper calls, and use `$WORKER_ISSUE_NUMBER` for direct issue access.

**Expected impact:** ~15-25K token savings and 5-8 fewer tool calls per worker session. Based on the batch analysis, workers that got past worktree setup made correct edits — this should significantly improve completion rates.

## Runtime Testing

- **Risk level:** Medium (dispatch infrastructure, no data deletion or auth changes)
- **Verification:** `shellcheck` clean on all 4 modified files. Changes are additive (env vars, early-exits, fast-paths) — existing interactive flows are unaffected. The worktree pre-creation falls back gracefully on failure.
- **Testing:** self-assessed — changes will be validated by the next pulse dispatch cycle

[...]

PR #18061, opened 2026-04-10
fix: run worktree pre-creation from repo_path, not pulse cwd

marcusquinn:
## Summary

Fix worktree pre-creation in `dispatch_with_dedup()` — it was always failing because `worktree-helper.sh` was called from the pulse-wrapper's cwd (`/` under launchd) instead of from within the repo directory.

**Root cause:** `worktree-helper.sh add` calls `git worktree add` which requires being inside a git repo. The pulse-wrapper runs as a launchd service with cwd `/`. The fix wraps the call in `cd "$repo_path" && ...`.

**Evidence:** Pulse log showed `Warning: worktree pre-creation failed for #18056 — worker will create its own` on the first dispatch after v3.6.225 deployed.

## Runtime Testing

- **Risk level:** Low (one-line cwd fix in subshell)
- **Verification:** shellcheck clean

Resolves the worktree pre-creation failure from v3.6.225 (#18046).

### DIFF
--- a/.agents/scripts/pulse-wrapper.sh
+++ b/.agents/scripts/pulse-wrapper.sh
@@ -8236,11 +8236,13 @@ dispatch_with_dedup() {
 	# existing path. On failure, fall back to letting the worker create it.
 	local worker_worktree_path="" worker_worktree_branch=""
 	local _wt_helper="${SCRIPT_DIR}/worktree-helper.sh"
-	if [[ -x "$_wt_helper" ]]; then
+	if [[ -x "$_wt_helper" && -d "$repo_path" ]]; then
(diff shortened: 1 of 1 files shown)


marcusquinn:
## Completion Summary

- **What**: Fix worktree pre-creation to run from repo_path instead of pulse cwd
- **Issue**: Worktree pre-creation from v3.6.225 always failed (not in git repo)
- **Files changed**: pulse-wrapper.sh
- **Testing**: shellcheck clean, root cause confirmed from pulse.log
- **Key decisions**: Used cd in subshell rather than git -C because worktree-helper.sh uses multiple git commands internally

Issue #18285, opened 2026-04-11
Bug: headless worker silently falls back to CLI when required MCP tools absent — causes infinite pulse dispatch loop


johnwaldo:
## Description

When a task body specifies an MCP-only approach (e.g. \"use shopify-dev-mcp only, do not use CLI\") and the required MCP server tools are not present in the headless worker's session tool list, the worker:

1. Correctly identifies the MCP is absent
2. Silently falls back to the prohibited CLI alternative
3. Completes partially or hits API errors
4. Exits via `process_exit` without posting `FULL_LOOP_COMPLETE`
5. Issue remains open — pulse re-dispatches — new worker repeats the cycle

This produces an infinite dispatch loop that runs until the pulse is manually disabled or the issue is force-closed.

## Steps to Reproduce

1. Configure a repo with `platform: "shopify"` in `repos.json` and `shopify-dev-mcp` in `~/.claude.json` (project-level, Claude Code)
2. Create an issue with task body that explicitly requires shopify-dev-mcp and forbids CLI fallback
3. Dispatch a headless worker (opencode runtime) via the pulse
4. Worker session does not have shopify-dev-mcp in its MCP tool list
5. Worker notes absence, falls back to `shopify store execute` CLI, hits partial errors, exits
6. Pulse re-dispatches — repeat

Observed: **183 dispatch cycles over ~16 hours** against a single issue. Each dispatch triggered a `shopify store auth` OAuth call (browser-based), resulting in ~40+ OAuth auth events logged on the host system.

## Expected Behavior

When a headless worker cannot find a required MCP server in its session tool list, it should:

- Exit `BLOCKED` immediately with reason: `"Required MCP server [name] not available in this session — cannot proceed without prohibited fallback"`
- NOT attempt CLI alternatives when the task body explicitly forbids them
- Leave an actionable BLOCKED comment so the dispatcher can route differently or update the issue

## Error Output

Worker reasoning from session log (paraphrased from actual output):

```
"I don't see any shopify-dev-mcp tools listed directly... 
The prohibition [against CLI] might be outdated or re-interpretable since MCP is unavailable."
```

Worker then proceeds with `shopify store execute`, hits GraphQL schema errors on article queries, exits without completing, issue stays open.

## Scope — Not Shopify-Specific

This affects any task that specifies an MCP-required approach. Current exposure on our system:

| Platform | Repos | MCP configured | At risk |
|----------|-------|---------------|---------|
[...]

PR #18330, opened 2026-04-11
GH#18285: fix shopify-dev-mcp enabled:true so server starts on OpenCode load

marcusquinn:
## Summary

Fixes the root cause of the user-reported failure: `@shopify` agent had `shopify-dev-mcp_*: true` in its tools but the MCP server was `enabled: false`, so the process never started and all tool calls would fail.

## The two-layer model (clarified)

| Layer | Field | Value | Effect |
|---|---|---|---|
| Process startup | `mcp.shopify-dev-mcp.enabled` | `true` | Server process starts when OpenCode loads |
| Global call gate | `tools.shopify-dev-mcp_*` | `false` | No agent can call the tools by default |
| Per-agent call gate | `agent.shopify.tools.shopify-dev-mcp_*` | `true` | Only `@shopify` agent can call the tools |

`enabled: false` means the process never starts — `tools_*: true` in an agent stub only gates call permission, not process startup. Both layers are needed.

## Changes

- `generate-runtime-config.sh`: pass `"enabled":true` in shopify-dev-mcp registration JSON
- `shopify.md`: correct Activation section — "server starts on load, tool calls gated per-agent"
- `shopify-dev-mcp-config.json.txt`: correct `enabled: false` → `true` in opencode config example, update notes

## Verification

After `aidevops update` + OpenCode restart:
```bash
# Server should be enabled:true
python3 -c "import json; d=json.load(open('/Users/\$USER/.config/opencode/opencode.json')); print(d['mcp']['shopify-dev-mcp']['enabled'])"
# Expected: True

# Global tools gate should still be false
python3 -c "import json; d=json.load(open('/Users/\$USER/.config/opencode/opencode.json')); print(d['tools']['shopify-dev-mcp_*'])"
# Expected: False
```

Resolves #18285

---

---
[aidevops.sh](https://aidevops.sh) v3.6.241 plugin for [OpenCode](https://opencode.ai) v1.4.3 with claude-sonnet-4-6 spent 4h 25m and 9,936 tokens on this as a headless worker.

## Summary by CodeRabbit

* **Configuration**
  * Shopify MCP server now automatically starts on system load with explicit enablement flag.
  * Tool access remains controlled at the agent level.

* **Documentation**
  * Updated Shopify agent documentation with revised setup instructions and clarified tool access behavior.

### DIFF
--- a/.agents/scripts/generate-runtime-config.sh
+++ b/.agents/scripts/generate-runtime-config.sh
@@ -993,12 +993,13 @@ _generate_mcp_for_runtime() {
 		'{"url":"https://mcp.cloudflare.com/mcp"}'
 	mcp_count=$((mcp_count + 1))
 
-	# Shopify Dev MCP (disabled by default; enabled per-agent via @shopify)
+	# Shopify Dev MCP — server starts on load (enabled:true); tool calls gated per-agent
+	# via tools: shopify-dev-mcp_*: true in the @shopify agent stub (false globally).
 	# Requires: Node 18+, Shopify CLI 3.93.0+ (npm install -g @shopify/cli@latest)
 	# TODO(permission-migration): when anomalyco/opencode#6892 is resolved, the
 	# per-agent tools: entry can be replaced with permission: shopify-dev-mcp: allow
 	register_mcp_for_runtime "$runtime_id" "shopify-dev-mcp" \
-		'{"command":"npx","args":["-y","@shopify/dev-mcp@latest"]}'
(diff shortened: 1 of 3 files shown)


marcusquinn:
Closing — root cause analysis was incomplete. The plugin's mcp-registry.mjs and agent-loader.mjs are the authoritative source, not opencode.json. enabled:false is correct (lazy-load on demand). The real fix is adding shopify to AGENT_MCP_TOOLS in agent-loader.mjs and shopify-dev-mcp to getMcpRegistry() in mcp-registry.mjs.

PR #18688, opened 2026-04-13
GH#18668: remove flock from pulse instance lock — mkdir-only is sufficient

### LINKED ISSUE #18668: bug(pulse): flock FD 9 inheritance regression — 9>&- fix is structurally incomplete, recovery mechanism has fuser parsing bug (Linux)
## Description

The flock FD 9 inheritance bug from GH#18094 is still occurring. Child processes spawned by git commands during pulse cycles inherit FD 9 and hold the flock indefinitely, deadlocking all subsequent pulse cycles. The recovery mechanism from GH#18141 (Layer 2 inode recovery) also fails to trigger due to a `fuser` output parsing bug.

This is the fourth occurrence of this root issue. Previous fix attempts:

- **GH#18094** (PR #18105): Added `python3 -c "import fcntl; fcntl.fcntl(9, fcntl.F_SETFD, fcntl.FD_CLOEXEC)"`. Ineffective — `fcntl(F_SETFD)` operates on the calling process's FD table, not the parent bash.
- **GH#18264** (PR #18309): Replaced python3 approach with `9>&-` appended to child-spawning commands. Correct mechanism but incomplete coverage.
- **GH#18141**: Added Layer 2 self-recovery (inode recreation after 3+ bounces from non-pulse holder). Works in isolation but has a parsing bug that prevents it from triggering in practice.

## Evidence

**Observed on a pulse-enabled repo (Linux, systemd timer):**

The `bd` binary (beads — a git merge driver) was spawned by a git operation during a pulse cycle. It inherited FD 9 and daemonised:

```
PID 298770, PPID 2247 (session manager — parent pulse exited, reparented)
FD: l-wx------ 9 -> /home/rob/.aidevops/logs/pulse-wrapper.lock
Process: /home/rob/go/bin/bd daemon start
Started: Mon Apr 13 12:04:10 2026
```

22 consecutive bounces, pulse completely locked out for ~50 minutes, zero self-recovery.

### Problem 1: `9>&-` coverage is structurally incomplete

The fix from GH#18264 appended `9>&-` to specific child-spawning commands. Current coverage across all `pulse-*.sh` modules:

- Git commands **with** `9>&-`: **3**
- Git commands **without** `9>&-`: **44**

Any of the 44 uncovered git commands can trigger a merge driver (like `bd`), which daemonises and holds the flock. The `9>&-` approach is a blocklist model — it requires every current and future child-spawning command to be individually annotated. This is structurally unsound because:

- New git calls added to pulse stages won't have `9>&-` unless the developer knows about this requirement
- Git merge drivers, hooks, and filters can all spawn background processes
- `git pull`, `git worktree add`, `git push` (with pre-push hooks) are all potential triggers
- The requirement is invisible — nothing in the code communicates "every child-spawning command must close FD 9"

### Problem 2: Recovery mechanism `fuser` parsing bug

`pulse-instance-lock.sh:94`:
```bash
flock_holder_pid=$(fuser "$LOCKFILE" 2>/dev/null | tr -d ' ')
```

During a bounce, the current pulse process also has FD 9 open (from `exec 9>"$LOCKFILE"` at `pulse-wrapper.sh:1110`, which runs BEFORE the `flock -n 9` attempt). So `fuser` returns multiple PIDs — the orphaned holder AND the current pulse process:

```
fuser output: "298770 302558"
After tr -d ' ': "298770302558"
```

Then `ps -p "298770302558"` fails (not a valid PID), `flock_holder_comm` is empty, and the recovery condition `[[ -n "$flock_holder_comm" ]]` evaluates false. The recovery block is never entered.

Log evidence:
```
flock secondary guard: held by PID 298770302558302995302997 (unknown), bounce 1
flock secondary guard: held by PID 298770302558302995302997 (unknown), bounce 22
```

Note `(unknown)` — the `ps` lookup failed on every bounce.

## Considerations

This is an **architectural problem**, not a missing annotation. The `9>&-` blocklist approach cannot scale. Some alternatives for the maintainers to evaluate:

1. **Subshell wrapper for all external commands** — run everything through a function that closes FD 9 before exec. Allowlist model instead of blocklist.
2. **Don't use flock** — the mkdir-based lock is already documented as the "primary atomic primitive" (`pulse-instance-lock.sh:48`). If flock is supplementary and causes recurring deadlocks, its cost may exceed its value on Linux.
3. **Use `/proc/self/fd/9` with `O_CLOEXEC` via a compiled helper** — a small C binary that calls `fcntl(9, F_SETFD, FD_CLOEXEC)` in the parent process (bash cannot do this natively). Would make FD 9 close-on-exec for all child processes without per-command annotation.
4. **Open FD 9 with bash `{fd}>` syntax** (bash 4.1+) and close/reopen around external commands — complex but keeps flock.

Separately, the `fuser` parsing bug should be fixed regardless of the architectural direction — extracting the first PID or filtering out `$$` would make the recovery mechanism functional.

## Platform

This is **Linux-specific**. macOS does not use the flock supplementary layer (`pulse-instance-lock.sh:88` gates on `command -v flock`). macOS relies on mkdir-only locking, which is not affected by FD inheritance.

## Environment

- aidevops version: 3.8.6
- OS: Ubuntu 24.04.4 LTS
- Shell: bash 5.2.21(1)-release
- gh CLI: 2.89.0

marcusquinn:
## Summary

Removes the supplementary `flock` layer from `acquire_instance_lock()` after four recurring deadlock incidents (GH#18094 → GH#18141 → GH#18264 → GH#18668) all traced to the same root cause: **bash has no built-in for `fcntl(F_SETFD, FD_CLOEXEC)`**.

## Root cause

When `exec 9>"$LOCKFILE"` opened FD 9 in the parent pulse process, that FD was inherited by every child: git commands, hooks, merge drivers, backgrounded workers. Any child that daemonised then held the flock indefinitely, deadlocking all subsequent pulse cycles.

Three targeted fixes each addressed a symptom:
- GH#18094: `python3 fcntl` — ineffective, operated on child's FD table not parent's
- GH#18141: inode self-recovery — `fuser` concatenated multiple PIDs; `ps` lookup failed
- GH#18264: `9>&-` annotation on child calls — correct mechanism but blocklist model (44 uncovered git calls at time of removal)

## Fix: Path A — remove flock entirely

`mkdir` atomicity is POSIX-guaranteed on all local filesystems (APFS, HFS+, ext4, btrfs, xfs). It is already documented as the "primary atomic primitive". The flock layer added defence against a scenario mkdir already handles, at the cost of recurring deadlocks.

## Files changed

- **`pulse-instance-lock.sh`**: remove 75-line flock section from `acquire_instance_lock()`, remove `exec 9>&-` from `release_instance_lock()`, update module header
- **`pulse-wrapper.sh`**: remove `exec 9>"$LOCKFILE"` setup, `LOCKFILE` variable, `9>&-` on worker launch, update lock protocol documentation
- **`pulse-logging.sh`**: remove `deadlock_*` fields from `pulse-health.json` schema
- **`pulse-ancillary-dispatch.sh`, `pulse-dispatch-core.sh`, `pulse-prefetch.sh`, `pulse-routines.sh`, `pulse-simplification.sh`, `pulse-watchdog.sh`**: remove all `9>&-` annotations (no longer needed)
- **`linters-local.sh`**: add `check_forbidden_exec_fd()` — categorical block on `exec N>*.lock` patterns in pulse scripts; enforces the policy going forward
- **`reference/bash-fd-locking.md`** (new): post-mortem and policy document; explains why re-introducing flock would fail

## Verification

```bash
shellcheck .agents/scripts/pulse-instance-lock.sh \
           .agents/scripts/pulse-wrapper.sh \
           .agents/scripts/pulse-logging.sh \
           .agents/scripts/linters-local.sh
# clean

.agents/scripts/linters-local.sh --gate forbidden-exec-fd
# 0 violations
```

Resolves #18668

### DIFF
--- a/.agents/reference/bash-fd-locking.md
+++ b/.agents/reference/bash-fd-locking.md
@@ -1,0 +1,58 @@
+# Bash FD Locking — Why flock was removed from the pulse instance lock
+
+## Summary
+
+The pulse instance lock uses **mkdir atomicity only**. The supplementary `flock`
+layer (FD 9) was removed in GH#18668 after four recurring deadlock incidents.
+
+## Background
+
+`pulse-instance-lock.sh` uses `mkdir "$LOCKDIR"` as the primary atomic lock
+primitive. Because mkdir is POSIX-guaranteed atomic on all local filesystems
+(APFS, HFS+, ext4, btrfs, xfs), it provides sufficient mutual exclusion without
+additional layers.
+
+A supplementary `flock` on a lock file (FD 9) was added in GH#4513 as a
+"belt-and-suspenders" guard for Linux. It was intended as defence-in-depth,
+not as the primary primitive.
+
+## The recurring problem: FD inheritance
+
+Bash has no built-in equivalent of `fcntl(F_SETFD, FD_CLOEXEC)`. When the pulse
+opens `exec 9>"$LOCKFILE"` in the parent bash process, FD 9 is inherited by
+**every child process** that bash forks — including:
+
+- git commands (`git pull`, `git push`, `git worktree add`, `git rebase`)
+- git hooks (pre-commit, post-receive, commit-msg)
+- git merge drivers (`bd`, custom merge tools)
+- all backgrounded workers
+
+Any of these can daemonise (reparent to PID 1), permanently holding the flock
+after the pulse exits. The next pulse cycle is then deadlocked indefinitely.
+
+## Four failed fix attempts
+
+| Issue | Fix attempted | Why it failed |
+|-------|--------------|---------------|
+| GH#18094 | `python3 -c "import fcntl; fcntl.fcntl(9, fcntl.F_SETFD, FD_CLOEXEC)"` | `fcntl(F_SETFD)` operates on the calling process's FD table. Running it in a child python3 process only set CLOEXEC on python's copy of FD 9, which was discarded when python exited. The parent bash FD 9 was never modified. |
+| GH#18141 | Layer 2 inode self-recovery after 3 bounces | Had a `fuser` parsing bug: `fuser` returned multiple PIDs (orphan + current pulse), `tr -d ' '` concatenated them, `ps -p "298770302558"` failed, `flock_holder_comm` was empty, recovery condition never triggered. |
+| GH#18264 | Append `9>&-` to all child-spawning commands | Correct mechanism but blocklist model. 44 uncovered git calls at time of removal. Any new git call without `9>&-` restores the vulnerability. Invisible requirement — nothing in the code communicates it. |
+| GH#18668 | This issue | Identified as architectural problem; adopted Path A (remove flock). |
+
+## Why mkdir-only is sufficient
+
+- mkdir is atomic at the kernel level. No TOCTOU race is possible.
+- Works on all local filesystems on all platforms — macOS and Linux.
+- No utility dependencies (unlike `flock` which requires util-linux on Linux).
+- Stale locks from SIGKILL or power loss are handled by PID-file staleness
+  detection in `acquire_instance_lock()`.
+- The `flock` layer added defence-in-depth against a scenario that mkdir already
+  handles: concurrent invocations. Its cost (deadlock risk) exceeded its value.
+
+## Policy
+
+- Do not re-add `flock` or any other persistent-FD locking in pulse scripts.
+- If additional concurrency protection is needed, use a separate lock file with
+  a short-lived FD (opened and closed within a single function call).
+- Reference this document when reviewing PRs that add `exec N>file` patterns
+  to pulse orchestration code.
--- a/.agents/scripts/linters-local.sh
+++ b/.agents/scripts/linters-local.sh
@@ -382,6 +382,81 @@ check_string_literals() {
 	return 0
 }
 
+# check_forbidden_exec_fd (GH#18668): categorical block on persistent FD-based
+# file locks in pulse-adjacent scripts. Matches any `exec N>PATH` where PATH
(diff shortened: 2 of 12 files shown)


marcusquinn:
Adding a follow-up commit with a 20-assertion characterization test at `.agents/scripts/tests/test-pulse-instance-lock-mkdir-only.sh`.

**Why**: the main commit removes the flock code; this test makes re-introducing it fail loudly. Seven scenarios:

1. Clean acquisition: `LOCKDIR` + PID file created, `_LOCK_OWNED` set
2. Stale lock recovery: dead PID cleared and re-acquired
3. Live concurrent owner: blocked, `LOCKDIR` preserved, `_LOCK_OWNED` unchanged
4. Release idempotency: `LOCKDIR` removed, safe to call twice
5. **No `LOCKFILE` created**: legacy `~/.aidevops/logs/pulse-wrapper.lock` must never appear during acquire or release — positive assertion that the file-based lock is gone
6. **FD 9 not open in parent**: negative test via `/dev/fd/9` probe — enforces that the persistent-FD inheritance vector is structurally removed. This is the assertion that would have caught GH#18094, GH#18141, and GH#18264 on day one.
7. Acquire/release cycle stability: 3 back-to-back cycles (the scenario that deadlocked the historical flock incidents) all succeed

The test sources `pulse-instance-lock.sh` directly with a minimal stub environment so it asserts the module's contract in isolation from the rest of `pulse-wrapper.sh`. Runs in ~250ms, no network, no GH state.

This closes the review recommendation I posted earlier — the "prevents re-regression" half of Path A.

PR #18793, opened 2026-04-14
GH#18786: fix set -e exit in interactive-session-helper.sh claim flow

### LINKED ISSUE #18786: t2075: fix(interactive-session-helper): `set -e` kills claim flow before `status:in-review` label is applied
## What

`interactive-session-helper.sh claim` silently dies with exit 1 and never applies `status:in-review` or self-assignment, breaking the t2056 interactive issue ownership protocol completely on every call. No output, no warning — the caller has no idea the claim didn't land.

## Why

Lines 265-271 of `.agents/scripts/interactive-session-helper.sh` carry the same `set -e` exit-propagation bug as GH#18770:

```bash
local had_label=1
_isc_has_in_review "$issue" "$slug"
local has_rc=$?
if [[ $has_rc -eq 0 ]]; then
    had_label=0
fi
```

`_isc_has_in_review` returns 1 as a legitimate "label not present" signal (jq `-e` on an empty match). Under `set -euo pipefail` at the top of the script, an unchecked non-zero return from a called function kills the parent function immediately — before `has_rc=$?` can capture it. The `_isc_cmd_claim` function exits with 1, the caller sees no output, and the issue never gets the `status:in-review` label.

**Reproduced on:** bash 3.2.57 (macOS), bash 4+ (Linux) — `set -e` behaviour is identical on both. Traced with `bash -x`:

```text
+ local had_label=1
+ _isc_has_in_review 18784 marcusquinn/aidevops
+ jq -e ...
+ return 1
(script dies here, `local has_rc=$?` never runs)
```

Impact: every interactive session that tries to claim an issue goes through this path. The pulse dispatch-dedup guard relies on `status:in-review` being set to block worker dispatch on an issue a human is actively working on (t2056). With this bug, that protection is silently disabled — the race between interactive claim and pulse dispatch is wide open.

This is the **same bug pattern** as GH#18770 (set -e + `local var=$(f)` capture) and GH#18784 (set -e + unguarded getent) — a systemic class. The new `bash-compat.md` pre-merge checklist item (from PR #18785) specifically calls this out as checklist item 4.

## How

Replace lines 265-271 with a `set -e`-safe idiom. Two equivalent fixes:

**Option A** (explicit rc capture):

```bash
local had_label=1
local has_rc=0
_isc_has_in_review "$issue" "$slug" || has_rc=$?
if [[ $has_rc -eq 0 ]]; then
    had_label=0
fi
```

**Option B** (direct conditional — cleaner):

```bash
local had_label=1
if _isc_has_in_review "$issue" "$slug"; then
    had_label=0
fi
```

Option B is preferred — it is shorter, idiomatic, and avoids the rc variable entirely. The function is already being called for its return code; there is no reason to split the call from the branch.

Also audit line 355 (`_isc_has_in_review "$issue" "$slug"` in `_isc_cmd_release`) for the same pattern. If it's a bare call followed by `$?` capture, apply the same fix.

## Acceptance criteria

- [ ] `interactive-session-helper.sh claim <N> <slug>` successfully applies `status:in-review` when the label is absent
- [ ] `interactive-session-helper.sh claim <N> <slug>` is idempotent when the label is already present (refreshes stamp, exits 0)
- [ ] No bare function calls followed by `$?` capture remain in `interactive-session-helper.sh`
- [ ] Regression test added covering the "label absent" path end-to-end with a stub `gh` and `set -e` enabled
- [ ] `bash -x interactive-session-helper.sh claim 18784 marcusquinn/aidevops` shows the full code path executing (not dying at `_isc_has_in_review`)

## Context

- Originating commit: 72f4bb1f2 (t2056, PR #18744) — the helper was brand new and the pattern was not reviewed for `set -e` propagation
- Sibling instance of the same bug class: GH#18770 (pulse-wrapper.sh self-check), GH#18784 (aidevops.sh getent)
- Pre-merge checklist that should have caught this: `.agents/reference/bash-compat.md` (from PR #18785) → "Cross-platform command portability" → item 4
- Affected protocol: t2056 "Interactive issue ownership" in `.agents/AGENTS.md` and `prompts/build.txt`

marcusquinn:
## Summary

Repair three latent bugs in interactive-session-helper.sh that combined to break the t2056 interactive issue ownership protocol. Under set -euo pipefail, the bare _isc_has_in_review call at line 267 propagated its legitimate 'label absent' return (rc=1) out of _isc_cmd_claim before 'has_rc=$?' could capture it, so the function exited silently and never applied status:in-review. The same pattern lived at line 355 in _isc_cmd_release. Both were replaced with set -e-safe idioms ('if f; then' in claim, '|| rc=$?' in release). A second latent bug in _isc_has_in_review ('jq any(.name; cond)' form) always raised 'Cannot index array with string' on jq 1.7+, masked by '2>&1 /dev/null', which made the idempotency branch unreachable even before set -e killed it. A third latent bug in _isc_cmd_release ('${extra_flags[@]}' under bash 3.2 set -u) was previously unreachable because of the jq bug; now that the path is live, it needed the standard '${arr[@]+"${arr[@]}"}' guard from reference/bash-compat.md. The main() dispatcher was also hardened so subcommand returns are captured with '|| rc=$?' rather than relying on 'return $?' after a case block that set -e would have killed.

## Files Changed

.agents/scripts/interactive-session-helper.sh,.agents/scripts/tests/test-interactive-session-claim.sh

## Runtime Testing

- **Risk level:** Low (agent prompts / infrastructure scripts)
- **Verification:** shellcheck clean on both files; all 18 tests pass (12 original + 6 new regression assertions); verified regression tests FAIL on pristine buggy version (5 distinct failures caught); end-to-end verified against GH#18786 via both idempotent and absent paths; bash -x trace confirms full code path executes to completion

Resolves #18786

---
[aidevops.sh](https://aidevops.sh) v3.8.15 plugin for [OpenCode](https://opencode.ai) v1.4.3 with claude-opus-4-6 spent 10m and 29,468 tokens on this as a headless worker.

### DIFF
--- a/.agents/scripts/interactive-session-helper.sh
+++ b/.agents/scripts/interactive-session-helper.sh
@@ -122,12 +122,23 @@ _isc_gh_reachable() {
 
 # Check whether an issue already carries `status:in-review`. Returns 0 when
 # present, 1 when absent, 2 when the metadata lookup failed.
+#
+# NOTE on the jq query: the two-argument form `any(generator; condition)`
+# passed `.name` as the generator on `(.labels // [])`, which tries to
+# index the *array itself* with string "name" and raises
+# "Cannot index array with string". The single-argument form
+# `any(condition)` iterates over the array automatically, which is both
+# shorter and correct. This was a latent bug that masked the bug fixed
+# in GH#18786 — the jq error (exit 5) was swallowed by `>/dev/null 2>&1`
+# and the function always returned 1 ("label absent"), so the idempotency
+# branch was dead code even before the set -e exit propagation killed
+# the whole claim flow.
 _isc_has_in_review() {
 	local issue="$1"
 	local slug="$2"
 	local json
 	json=$(gh issue view "$issue" --repo "$slug" --json labels 2>/dev/null) || return 2
-	if printf '%s' "$json" | jq -e '(.labels // []) | any(.name; . == "status:in-review")' >/dev/null 2>&1; then
+	if printf '%s' "$json" | jq -e '(.labels // []) | any(.name == "status:in-review")' >/dev/null 2>&1; then
 		return 0
 	fi
 	return 1
@@ -262,15 +273,15 @@ _isc_cmd_claim() {
 		return 0
 	fi
 
-	# Idempotency: if already in-review, just refresh the stamp and exit
-	local had_label=1
-	_isc_has_in_review "$issue" "$slug"
-	local has_rc=$?
-	if [[ $has_rc -eq 0 ]]; then
-		had_label=0
-	fi
-
-	if [[ $had_label -eq 0 ]]; then
+	# Idempotency: if already in-review, just refresh the stamp and exit.
+	# NOTE: `_isc_has_in_review` legitimately returns non-zero ("label absent"
+	# = 1, "lookup failed" = 2). Under `set -e`, a bare call followed by
+	# `local rc=$?` capture kills this function before `rc=$?` runs — because
+	# the unchecked non-zero return propagates up through the parent function.
+	# Use a direct `if` conditional so the return is consumed by the branch,
+	# which `set -e` treats as a tested condition and does not propagate.
+	# Sibling bug class: GH#18770 (pulse self-check), GH#18784 (aidevops.sh getent).
+	if _isc_has_in_review "$issue" "$slug"; then
 		_isc_info "claim: #$issue already has status:in-review — refreshing stamp"
 		_isc_write_stamp "$issue" "$slug" "$worktree_path" "$user"
 		return 0
@@ -350,10 +361,14 @@ _isc_cmd_release() {
 		return 0
 	fi
 
-	# Idempotency: skip label work if not in-review
-	local has_rc
-	_isc_has_in_review "$issue" "$slug"
-	has_rc=$?
+	# Idempotency: skip label work if not in-review. `_isc_has_in_review`
+	# has three return states (0 = present, 1 = absent, 2 = lookup failed),
+	# so we need the actual rc — but a bare call under `set -e` propagates
+	# non-zero returns before `rc=$?` can capture them. Use `|| rc=$?` which
+	# is a tested condition that `set -e` does not propagate. Default to 0
+	# so the "present" branch (rc=0) falls through to the transition below.
+	local has_rc=0
+	_isc_has_in_review "$issue" "$slug" || has_rc=$?
 	if [[ $has_rc -eq 1 ]]; then
 		_isc_info "release: #$issue not in status:in-review — no-op"
 		return 0
@@ -363,7 +378,12 @@ _isc_cmd_release() {
 		return 0
 	fi
 
-	# Transition in-review -> available
+	# Transition in-review -> available. Build the flag list as a plain
+	# array and expand it with the `${arr[@]+"${arr[@]}"}` guard so an
+	# empty array doesn't trip `set -u` on bash 3.2 (macOS default). This
+	# is the idiom documented in reference/bash-compat.md "empty array
+	# expansion". Fourth latent bug found alongside GH#18786: previously
+	# masked because the broken jq query made this branch unreachable.
 	local -a extra_flags=()
 	if [[ $unassign -eq 1 ]]; then
 		local user
@@ -373,7 +393,7 @@ _isc_cmd_release() {
 		fi
 	fi
 
-	if set_issue_status "$issue" "$slug" "available" "${extra_flags[@]}" >/dev/null 2>&1; then
+	if set_issue_status "$issue" "$slug" "available" ${extra_flags[@]+"${extra_flags[@]}"} >/dev/null 2>&1; then
 		_isc_info "release: #$issue → status:available"
 		return 0
 	fi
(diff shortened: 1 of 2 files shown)


marcusquinn:
## Completion Summary

- **What**: Repair three latent bugs in interactive-session-helper.sh that combined to break the t2056 interactive issue ownership protocol. Under set -euo pipefail, the bare _isc_has_in_review call at line 267 propagated its legitimate 'label absent' return (rc=1) out of _isc_cmd_claim before 'has_rc=$?' could capture it, so the function exited silently and never applied status:in-review. The same pattern lived at line 355 in _isc_cmd_release. Both were replaced with set -e-safe idioms ('if f; then' in claim, '|| rc=$?' in release). A second latent bug in _isc_has_in_review ('jq any(.name; cond)' form) always raised 'Cannot index array with string' on jq 1.7+, masked by '2>&1 /dev/null', which made the idempotency branch unreachable even before set -e killed it. A third latent bug in _isc_cmd_release ('${extra_flags[@]}' under bash 3.2 set -u) was previously unreachable because of the jq bug; now that the path is live, it needed the standard '${arr[@]+"${arr[@]}"}' guard from reference/bash-compat.md. The main() dispatcher was also hardened so subcommand returns are captured with '|| rc=$?' rather than relying on 'return $?' after a case block that set -e would have killed.
- **Issue**: #18786
- **Files changed**: .agents/scripts/interactive-session-helper.sh,.agents/scripts/tests/test-interactive-session-claim.sh
- **Testing**: shellcheck clean on both files; all 18 tests pass (12 original + 6 new regression assertions); verified regression tests FAIL on pristine buggy version (5 distinct failures caught); end-to-end verified against GH#18786 via both idempotent and absent paths; bash -x trace confirms full code path executes to completion
- **Key decisions**: Chose 'if f; then' (option B) for _isc_cmd_claim idempotency check per issue's preferred option — shorter, idiomatic, avoids rc variable entirely. Used '|| rc=$?' (option A) for _isc_cmd_release because the function has three meaningful return states (0/1/2) that a simple boolean branch cannot express. Extended scope to fix the latent jq query and bash 3.2 array expansion bugs because the primary set -e fix reroutes execution through both of them; leaving either in place would mean the fix doesn't actually restore working behaviour. Audited the main() dispatcher for the same class and hardened it to capture subcommand returns via '|| rc=$?'. Regression tests execute the script as a subprocess (not sourced) so set -euo pipefail is live — without this, the existing sourced tests mask the entire bug class (documented inline in the test).

marcusquinn:
## Resolved via PR #18793 → v3.8.16

### What was done

Repaired three latent bugs in `interactive-session-helper.sh` that combined to silently break the t2056 interactive issue ownership protocol on every call:

1. **Primary (reported): `set -e` exit propagation in `_isc_cmd_claim`** — the bare `_isc_has_in_review` call at line 267 propagated its legitimate "label absent" return (rc=1) out of the parent function before `local has_rc=$?` could capture it, exiting silently without applying `status:in-review`. Fixed with the `if f; then` idiom (option B from the issue). Same pattern at line 355 in `_isc_cmd_release` fixed with `|| rc=$?` (option A — needed because release distinguishes three return states 0/1/2).
2. **Latent: broken `jq` query in `_isc_has_in_review`** — `any(.name; cond)` raised `Cannot index array with string "name"` on jq 1.7+, swallowed by `2>&1 /dev/null`, so the function always returned "label absent". This made the idempotency branch dead code *even without* the set -e bug — every repeat claim would have re-triggered a full label transition. Replaced with the single-argument `any(cond)` form.
3. **Latent: bash 3.2 empty-array expansion under `set -u`** — `"${extra_flags[@]}"` in `_isc_cmd_release` tripped "unbound variable" on macOS default bash. Previously unreachable because of bug #2; now that the path is live, fixed with the `${arr[@]+"${arr[@]}"}` guard documented in `reference/bash-compat.md`.

Also hardened `main()` so subcommand returns are captured with `|| rc=$?` rather than relying on `return $?` after a `case` block that `set -e` would have killed on any non-zero subcommand return (e.g. `_isc_cmd_status` returning 1 for "target issue not found").

### Testing evidence

- **Regression test (new)**: 6 assertions in `test-interactive-session-claim.sh` that execute the helper as a subprocess (not sourced) so `set -euo pipefail` is fully active — sourced tests mask the entire bug class.
- **Verified regression test catches pristine bug**: 5/6 new assertions fail on the pre-fix code; 18/18 pass after fix.
- **Live E2E**: `bash -x` trace confirms the full code path executes on the real issue (both label-absent and label-present paths). Deployed helper at `~/.aidevops/agents/scripts/` also passes.
- **Sibling linters**: `shellcheck` clean on both modified files.

### Key decisions

- **Scope expansion**: the reported bug was only #1, but fixing #1 in isolation would have left `_isc_cmd_claim` still broken because #2 made the idempotency branch unreachable and #3 would have crashed the newly-live release path on bash 3.2. Three bugs; one fix.
- **Idiom choice**: `if f; then` (cleaner) for the two-state claim check, `|| rc=$?` (preserves rc) for the three-state release check, `|| rc=$?` for the main dispatcher.
- **Regression test runs as subprocess**: without this, `set -euo pipefail` is masked by the test's `set +e` and the bug class is invisible. Documented inline so future tests know why.

### Files changed

- `.agents/scripts/interactive-session-helper.sh` — three fixes + hardened main() dispatcher
- `.agents/scripts/tests/test-interactive-session-claim.sh` — 6 new GH#18786 regression assertions + `export STUB_LOG` for subprocess stub visibility

### Released in

v3.8.16 — https://github.com/marcusquinn/aidevops/releases/tag/v3.8.16

### Follow-up

None. Acceptance criteria all met:

- [x] `claim` successfully applies `status:in-review` when absent (regression test case b)
- [x] `claim` is idempotent when label already present (live test + regression case d)
- [x] No bare function calls followed by `$?` capture remain in `interactive-session-helper.sh`
- [x] Regression test added covering "label absent" path end-to-end with `set -e` enabled
- [x] `bash -x` shows full code path executing (not dying at `_isc_has_in_review`)

---



---
[aidevops.sh](https://aidevops.sh) v3.8.16 plugin for [OpenCode](https://opencode.ai) v1.4.3 with claude-opus-4-6 spent 16m and 33,195 tokens on this as a headless worker. Solved in 45m.

PR #21538, opened 2026-04-28
GH#21440: fix _dt_json_field to use grep -oE instead of greedy sed

### LINKED ISSUE #21440: Review followup: PR #21435 — t3003: adaptive per-candidate dispatch timeout
## Unaddressed review bot suggestions

PR #21435 was merged with unaddressed review bot feedback. Each comment
below includes its file path, line number, a direct link to the inline
review comment, and a `diff` fence with the code context the bot was
flagging. Resolved and outdated threads are filtered out via GitHub's
GraphQL review-thread state. Read the relevant lines, decide whether
the suggestion is correct, and either apply the fix or close this issue
with a wontfix rationale.

**Source PR:** https://github.com/marcusquinn/aidevops/pull/21435

---

### You are the triager (worker-is-triager rule)

This issue is **auto-created from review bot output** and dispatched
directly to you. Review bots can be wrong: hallucinated line refs, false
premises about codebase structure, template-driven sweeps without
measurements (see GH#17832-17835 for prior art and `prompts/build.txt`
section 6a). **Do not assume the bot is correct.** Verify before acting.

You must end in exactly one of three outcomes — no fourth "hand it back
to the human" path exists. Humans approve decisions; they do not re-do
analysis.

#### Outcome A — Premise falsified → close the issue

1. Read the cited `file:line` (listed under *Files to modify* below).
2. If the bot's claim is factually wrong (file doesn't exist at that
   line, function doesn't behave as described, "auto-generated" section
   isn't actually auto-generated, etc.), **close the issue** with a
   comment in this shape:

   > **Premise falsified.** \<what the bot claimed\>. \<what the code
   > actually shows, with a `file:line` citation or one-line quote\>.
   > Not acting.

   No PR. No further dispatch. The closing comment trains the next
   session reading this thread and the noise filter.

#### Outcome B — Premise correct + fix is obvious → implement and PR

1. Verify the bot's premise as above.
2. Read the Worker Guidance section below, open a worktree, implement.
3. Open a PR with `Resolves #<this-issue-number>` in the body
   (use THIS issue's number, not the source PR's) so merge auto-closes it.
4. Follow the normal Lifecycle Gate (brief, tests, review-bot-gate,
   merge, postflight).

#### Outcome C — Premise correct but approach is a genuine judgment call

Only use this path if you reach it after Outcomes A and B don't apply:
the bot's finding is real, but the fix requires a decision that is
architectural, policy, breaking-change, or otherwise genuinely outside
what you can resolve autonomously. In that case, post a **decision
comment** with exactly these fields:

- **Premise check:** one line, confirming the finding is real.
- **Analysis:** 2-4 bullets on the trade-offs.
- **Recommended path:** the option you would take if the decision were
  yours, with rationale.
- **Specific question:** the single decision the human needs to make
  (yes/no or pick-one, not open-ended).

Then apply `needs-maintainer-review` and stop. The human wakes up to a
ready-to-approve recommendation, not a blank task.

> **Ambiguity about scope or style is not Outcome C.** Per
> `prompts/build.txt` "Reasoning responsibility", the model does the
> thinking and delivers a recommendation. Only escalate what is genuinely
> a maintainer-only decision.

### Worker Guidance

**Files to modify:**

- `.agents/scripts/dispatch-timing-helper.sh:301`

**Implementation steps (Outcome B path):**

1. Read the `diff` block under each inline comment below — it shows the
   exact code the bot was flagging. Open the file only if you need
   surrounding context beyond what the diff tail shows.
2. Read the bot's full comment below the diff — it contains the rationale
   and any suggested change.
3. Verify the premise before implementing (see Outcome A). If the premise
   is wrong, switch to Outcome A instead of burning iterations trying to
   satisfy a wrong suggestion.
4. If multiple comments target the same file, group your edits into one
   logical commit.
5. Run `shellcheck` / `markdownlint-cli2` / project tests as appropriate.

**Verification:**

- Open the new PR with `Resolves #<this-issue>` so this followup is auto-closed on merge.
- If the bot's suggestion was incorrect, close this issue with a Outcome A comment — do not open a no-op PR.

#### gemini-code-assist on `.agents/scripts/dispatch-timing-helper.sh:301`

[View inline comment](https://github.com/marcusquinn/aidevops/pull/21435#discussion_r3150918546)

```diff
+	# Match "field":"value" (string) OR "field":NUMBER (numeric)
+	# Try string first
+	local val
+	val=$(printf '%s' "$line" | sed -nE "s/.*\"${field}\":\"([^\"]*)\".*/\1/p")
+	if [[ -n "$val" ]]; then
+		echo "$val"
+		return 0
+	fi
+	# Try numeric
+	val=$(printf '%s' "$line" | sed -nE "s/.*\"${field}\":([0-9]+).*/\1/p")
+	echo "$val"
+	return 0
```

> high
> 
> The regex-based extraction in `_dt_json_field` uses a greedy `.*` prefix which can lead to incorrect value extraction if a preceding field's value contains the target field name as a substring. For example, if a `repo` value is `"my-outcome-repo"`, extracting the `outcome` field will incorrectly match the substring within the `repo` value instead of the actual `outcome` field. 
> 
> Using `grep -oE` to isolate the specific key-value pair is a more robust approach that avoids greedy matching issues while remaining compatible with Bash 3.2 and avoiding a `jq` dependency. Additionally, per repository rules, when extracting metadata fields from a line using `grep`, use `tail -1` instead of `head -1` to ensure robustness, especially when the field might appear multiple times.
> 
> ```suggestion
> 	# Try string first
> 	local val
> 	val=$(printf '%s' "$line" | grep -oE '"'"${field}"'"[[:space:]]*:[[:space:]]*"[^"]*"' | tail -1 | cut -d'"' -f4)
> 	if [[ -n "$val" ]]; then
> 		echo "$val"
> 		return 0
> 	fi
> 	# Try numeric
> 	val=$(printf '%s' "$line" | grep -oE '"'"${field}"'"[[:space:]]*:[[:space:]]*[0-9]+' | tail -1 | cut -d: -f2 | tr -d ' ')
> 	echo "$val"
> ```
> 
> <details>
> <summary>References</summary>
> 
> 1. For shell scripts that need to be portable, prefer built-in tools like grep over external dependencies like jq for simple tasks, to avoid installation requirements on diverse systems.
> 2. When extracting metadata fields from a line using 'grep' and 'sed' in shell scripts, use 'tail -1' instead of 'head -1' to ensure robustness, especially when the field might appear multiple times or be at the end of the line.
> </details>

#### gemini-code-assist[bot] review summary

[View review](https://github.com/marcusquinn/aidevops/pull/21435#pullrequestreview-4185003887)

> ## Code Review
> 
> This pull request introduces an adaptive timeout mechanism for dispatch attempts, replacing fixed durations with dynamic budgets calculated from a rolling window of recent successes using EWMA and p95 metrics. The implementation includes a new `dispatch-timing-helper.sh` script, integration into the `pulse-dispatch-engine.sh`, and a suite of unit tests. Feedback focuses on improving the robustness of the regex-based JSON field extraction to avoid incorrect matches caused by greedy patterns.

---
[aidevops.sh](https://aidevops.sh) v3.13.5 automated scan.

marcusquinn:
## Summary

Replace the greedy `sed` extraction in `_dt_json_field` with `grep -oE`, addressing the Gemini review feedback on PR #21435.

## What changed

- `.agents/scripts/dispatch-timing-helper.sh`: `_dt_json_field` now uses `grep -oE '"field"[[:space:]]*:[[:space:]]*...'` instead of `sed -nE 's/.*"field":...'`
- Added `tail -1` to handle the duplicate-key edge case (returns last match, consistent with previous greedy-right behaviour)
- Comment updated to explain the rationale

## Why

The greedy `.*` prefix in the `sed` substitution pattern can theoretically match a field name that appears as an exact substring within a preceding field's value. The `grep -oE` approach isolates only the specific `"fieldname":value` token, making the extraction robust regardless of surrounding field values.

## Testing

- All 15 applicable unit tests pass (`test-dispatch-timing-helper.sh`; 3 pre-existing failures due to `mapfile` bash 4+ incompatibility in Test 13, unrelated to this change)
- `shellcheck` passes with zero violations

Resolves #21440

---
[aidevops.sh](https://aidevops.sh) v3.13.5 plugin for [OpenCode](https://opencode.ai) v1.14.28 with claude-sonnet-4-6 spent 10m and 17,502 tokens on this as a headless worker.

### DIFF
--- a/.agents/scripts/dispatch-timing-helper.sh
+++ b/.agents/scripts/dispatch-timing-helper.sh
@@ -284,22 +284,27 @@ _dt_cmd_recommend() {
 }
 
 # Extract value of $field from a single JSONL record on stdin.
-# Uses a simple regex-based extractor — avoids jq for hot-path speed.
+# Uses grep -oE to isolate the specific key-value pair, which avoids the
+# greedy-prefix issue where sed's .* can match a field name that appears as
+# a substring inside a preceding field's value.  Bash 3.2 compatible — no
+# jq dependency.
 # Args: $1 = field name (string-valued or numeric).
 _dt_json_field() {
 	local field="$1"
 	local line
 	IFS= read -r line
-	# Match "field":"value" (string) OR "field":NUMBER (numeric)
+	# Match "field":"value" (string) OR "field":NUMBER (numeric).
+	# grep -oE emits only the matching text; tail -1 handles the rare
+	# duplicate-key edge case by preferring the last occurrence.
 	# Try string first
 	local val
-	val=$(printf '%s' "$line" | sed -nE "s/.*\"${field}\":\"([^\"]*)\".*/\1/p")
+	val=$(printf '%s' "$line" | grep -oE "\"${field}\"[[:space:]]*:[[:space:]]*\"[^\"]*\"" | tail -1 | cut -d'"' -f4)
 	if [[ -n "$val" ]]; then
 		echo "$val"
 		return 0
 	fi
 	# Try numeric
-	val=$(printf '%s' "$line" | sed -nE "s/.*\"${field}\":([0-9]+).*/\1/p")
+	val=$(printf '%s' "$line" | grep -oE "\"${field}\"[[:space:]]*:[[:space:]]*[0-9]+" | tail -1 | cut -d: -f2 | tr -d ' ')
 	echo "$val"
 	return 0
 }


marcusquinn:
## MERGE_SUMMARY

**PR #21538** — GH#21440: fix `_dt_json_field` to use `grep -oE` instead of greedy sed

### What was done

Replaced the `sed -nE "s/.*\"${field}\":..."` extraction in `_dt_json_field` (`.agents/scripts/dispatch-timing-helper.sh`) with `grep -oE "\"${field}\"[[:space:]]*:[[:space:]]*..."` to avoid greedy-prefix matching issues. Added `tail -1` to handle the duplicate-key edge case.

### Triage outcome

**Outcome B** (premise correct, fix implemented). The bot's specific example (`my-outcome-repo` as repo value) does not actually trigger a wrong match in the current JSONL format, but the `grep -oE` approach is genuinely more robust and avoids the theoretical concern about greedy `.*` matching field names appearing as substrings in preceding field values.

### Testing

- All 15 applicable unit tests pass (3 pre-existing failures in Test 13 due to `mapfile` bash 4+ incompatibility, unrelated to this change)
- `shellcheck` zero violations
- All CI checks passed

Resolves #21440


---
[aidevops.sh](https://aidevops.sh) v3.13.5 plugin for [OpenCode](https://opencode.ai) v1.14.28 with claude-sonnet-4-6 spent 26m and 24,909 tokens on this as a headless worker.

PR #21764, opened 2026-04-29
GH#21729: fix _dff_dispatch_loop_parallel wait -n tight loop that grows pulse-wrapper.log to 679GB

### LINKED ISSUE #21729: bug(pulse): _dff_dispatch_loop_parallel wait -n tight loop grows pulse-wrapper.log to 679GB
## Description

`pulse-dispatch-engine.sh` line 969 has a `wait -n 2>/dev/null || true` call inside the parallel dispatch loop that spins in a tight infinite loop when a worker PID dies but isn't properly reaped from the `_pids` array. The error `wait: pid N is not a child of this shell` repeats millions of times per minute, growing `pulse-wrapper.log` to 679GB and consuming 73% of a 1TB drive.

## Expected Behavior

The parallel dispatch loop should properly reap dead PIDs from `_pids` before calling `wait -n`, and the wait loop should have a sleep guard to prevent tight spinning even if reaping fails.

## Steps to Reproduce

1. Run pulse with parallel dispatch enabled
2. A worker PID dies (e.g., watchdog kill, crash, or normal exit)
3. The PID is not removed from `_pids` array before the next `wait -n` call
4. `wait -n` fails with "pid N is not a child of this shell"
5. The loop retries immediately with no sleep, writing millions of error lines to the log

## Root Cause

In `_dff_dispatch_loop_parallel()` (line 968-975):

```bash
while ((${#_pids[@]} >= max_parallel)); do
    wait -n 2>/dev/null || true
    _alive_pids=()
    while IFS= read -r pid; do
        [[ -n "$pid" ]] && _alive_pids+=("$pid")
    done < <(_dff_reap_pids "${_pids[@]+${_pids[@]}}")
    _pids=("${_alive_pids[@]+${_alive_pids[@]}}")
done
```

The `wait -n` call fails on a dead PID, but `_pids` is only updated *after* the `wait -n` call. If `_dff_reap_pids` doesn't return the dead PID (e.g., it was already reaped by the kernel), the dead PID stays in `_pids` and `wait -n` fails again on the next iteration — infinitely, with no sleep.

The `2>/dev/null` redirect is also ineffective because the pulse wrapper redirects all stderr to the log file via `>>"$LOGFILE" 2>&1` at the invocation level.

## Evidence

```
$ ls -lhS ~/.aidevops/logs/pulse-wrapper.log
-rw-r--r--  1 justin  staff   679G Apr 29 06:26 pulse-wrapper.log

$ tail -100 ~/.aidevops/logs/pulse-wrapper.log | head -5
/Users/justin/.aidevops/agents/scripts/pulse-dispatch-engine.sh: line 873: wait: pid 22801 is not a child of this shell
/Users/justin/.aidevops/agents/scripts/pulse-dispatch-engine.sh: line 873: wait: pid 22801 is not a child of this shell
...

$ tail -10000 ~/.aidevops/logs/pulse-wrapper.log | sort | uniq -c | sort -rn | head -3
9993 /Users/justin/.aidevops/agents/scripts/pulse-dispatch-engine.sh: line 873: wait: pid 22801 is not a child of this shell
```

## Environment

- **aidevops version**: 3.13.10
- **AI Assistant**: OpenCode
- **OS**: macOS 26.4
- **Shell**: bash 5.3.9(1)-release
- **gh CLI**: gh version 2.92.0 (2026-04-28)

## Suggested Fix

1. Move the PID reaping *before* the `wait -n` call, or use `wait -n` only on PIDs that are confirmed alive
2. Add a `sleep 1` (or smaller) guard inside the wait loop to prevent tight spinning
3. Consider using `kill -0 "$pid" 2>/dev/null` to check PID liveness before calling `wait`

---
[aidevops.sh](https://aidevops.sh) v3.13.10 plugin for [OpenCode](https://opencode.ai) v1.14.29 with claude-opus-4-7 spent 9m and 9,371 tokens on this with the user in an interactive session.

alex-solovyev:
## Problem

`_dff_dispatch_loop_parallel()` in `pulse-dispatch-engine.sh` has a `wait -n` call inside the parallel dispatch concurrency-cap loop that spins in a tight infinite loop when a worker PID dies and the OS recycles the PID number for a different process.

**Root cause:** `kill -0 $pid` succeeds because the recycled PID exists in the process table, so `_dff_reap_pids` keeps returning it as "alive". But `wait -n` fails with "pid N is not a child of this shell" because the recycled PID belongs to a different process. The old code swallowed the error with `|| true` and looped back immediately — no sleep, no progress. This wrote millions of error lines per minute, growing `pulse-wrapper.log` to 679GB (73% of a 1TB drive).

## Fix

Three changes to the wait loop (lines 968-975):

1. **Reap before wait:** Move the `_dff_reap_pids` call above `wait -n` so dead PIDs (those where `kill -0` fails) are removed before we try to block.
2. **Re-check after reap:** If reaping dropped the count below `max_parallel`, break out of the while loop immediately — no need to call `wait -n`.
3. **Purge on wait -n failure:** When `wait -n` fails (returns non-zero), ALL remaining PIDs in `_pids` are stale (PID reuse). Purge the entire array, log the event, and sleep 1s to prevent any residual tight looping.

## Verification

- `shellcheck .agents/scripts/pulse-dispatch-engine.sh` — clean, no new violations
- Pre-commit hooks pass (ratchet checks, return statements, positional params)
- The fix is conservative: normal operation (valid child PIDs) is unchanged. The new code path only fires when `wait -n` reports no valid children, which is the exact error condition from the bug report.

Resolves #21729

---
[aidevops.sh](https://aidevops.sh) v3.13.10 plugin for [OpenCode](https://opencode.ai) v1.14.29 with claude-opus-4-6 spent 4m and 9,924 tokens on this as a headless worker.


alex-solovyev:
## Summary

Fixed the `wait -n` tight loop in `_dff_dispatch_loop_parallel()` that grew `pulse-wrapper.log` to 679GB. Root cause was PID reuse by the OS — `kill -0` reported recycled PIDs as alive, but `wait -n` failed on them ("not a child of this shell"), causing millions of iterations per minute with no sleep.

## Changes

- **`pulse-dispatch-engine.sh:968-993`**: Restructured the concurrency-cap wait loop to reap PIDs before calling `wait -n`, re-check the count after reaping, and purge all stale PIDs when `wait -n` fails (PID reuse scenario). Added 1s sleep guard on the purge path.

## Testing

- `shellcheck .agents/scripts/pulse-dispatch-engine.sh` — clean
- Pre-commit hooks pass (ratchet, return statements, positional params, string literals)
- Fix is conservative: normal code path unchanged; new path only fires on `wait -n` failure


---
[aidevops.sh](https://aidevops.sh) v3.13.10 plugin for [OpenCode](https://opencode.ai) v1.14.29 with claude-opus-4-6 spent 5m and 10,858 tokens on this as a headless worker. Overall, 6h 22m since this issue was created.

marcusquinn (approved):
Approved by maintainer @marcusquinn — wait -n tight-loop fix is independent.

PR #21951, opened 2026-04-30
GH#21944: lazy-start Claude CLI proxy + skip in headless + EADDRINUSE adopt

### LINKED ISSUE #21944: t3212: claude-proxy adopt-on-EADDRINUSE so multi-instance startup stops printing scary error
## Task

Stop starting the Claude CLI proxy unconditionally on every OpenCode plugin init. Lazy-start it only when a `claudecli/*` model is actually selected, and skip it entirely in headless workers (which never use claudecli/* models). Keep the EADDRINUSE adopt-on-retry as defense in depth for the rare case where two interactive sessions both legitimately need the proxy and race on bind.

## Why

Every OpenCode startup on a host running >=2 OpenCode instances prints:

```
[aidevops] Claude proxy: failed to start: Failed to start server. Is port 32125 in use?
```

Reproduced 2026-04-30 with 16 concurrent OpenCode instances on the user's host (1 interactive + 15 headless pulse workers). Only one binds 32125; the other 15 race and lose. The error is cosmetic (the proxy is up; subsequent OpenCode launches would adopt it if the probe succeeded), but it appears on every startup and erodes trust in pulse-worker telemetry.

Three independent failures stack:

1. Hardcoded port 32125 (claude-proxy.mjs:44) — every instance races for it.
2. Probe-and-adopt uses 1s timeout (claude-proxy.mjs:318) — too short when the existing proxy is mid-SSE-stream. Verified: `curl --max-time 2 http://127.0.0.1:32125/v1/models` timed out against the bound port.
3. `Bun.serve` EADDRINUSE catch logs and gives up (claude-proxy.mjs:388-394) — no re-probe, even though EADDRINUSE proves a proxy exists to adopt.

But the **root architectural issue** is upstream of the race: the proxy is starting on every plugin init, even though most sessions never select a `claudecli/*` model. Pulse workers always run with `-m anthropic/claude-X` (an Anthropic API model passed via cmdline) — they NEVER need the local Claude CLI proxy. Sibling proxies already gate on credentials:

- `cursor-proxy.mjs`: only starts if `getAccounts("cursor").length > 0` (index.mjs:151-161)
- `google-proxy.mjs`: only starts if `getAccounts("google").length > 0` (index.mjs:163-177)
- `claude-proxy.mjs`: NO gate — starts always when `claude` CLI binary is installed (index.mjs:179-187)

Lazy-load fixes the systemic cause, not just the symptom. With it, only the user's interactive session starts the proxy; the 15 pulse workers don't even try, so there's nothing to race.

### Files Scope

- `EDIT: .agents/plugins/opencode-aidevops/claude-proxy.mjs:44, 315-396`
- `EDIT: .agents/plugins/opencode-aidevops/index.mjs:179-187`
- `EDIT: .agents/plugins/opencode-aidevops/config-hook.mjs` (if provider registration needs to move out of `startClaudeProxy` so the model still appears in the picker without starting the server)

### Reference pattern

Three layered fixes; each is independently valuable, all three together is the systemic answer.

**(1) Skip proxy startup in headless workers (PRIMARY — biggest impact).** In `index.mjs:179-187`, before calling `startClaudeProxy`, check for headless env vars and short-circuit:

```js
const isHeadless = process.env.AIDEVOPS_HEADLESS === "1"
  || process.env.OPENCODE_HEADLESS === "1"
  || process.env.FULL_LOOP_HEADLESS === "1"
  || process.env.GITHUB_ACTIONS === "true";

if (!isHeadless) {
  try {
    const claudeProxyResult = await startClaudeProxy(client, directory);
    if (claudeProxyResult) {
      console.error(`[aidevops] Claude proxy started on port ${claudeProxyResult.port} with ${claudeProxyResult.models.length} models`);
    }
  } catch (err) {
    console.error(`[aidevops] Claude proxy failed to start: ${err.message}`);
  }
}
```

This alone eliminates the user-visible error in the canonical case (interactive + N pulse workers).

**(2) Lazy-start on first claudecli/* request (SECONDARY — addresses the genuine multi-interactive case).** Split `startClaudeProxy` into two phases:

- **Phase 1 (init-time, cheap)**: Register the `claudecli` provider entry in `opencode.json` so the models appear in the picker. Do NOT call `Bun.serve`. Move provider registration from `launchProxyServer` into a separate `registerClaudeProviderOnly()` function called from `index.mjs` config hook.
- **Phase 2 (on-demand)**: Add an `experimental.chat.messages.transform` (or similar pre-request) hook that inspects the active model. If it's `claudecli/*` and the proxy isn't started, call `ensureClaudeProxyStarted()` which does the `Bun.serve` + EADDRINUSE retry. Block the request until the proxy is up (max ~5s).

Alternative if `chat.messages.transform` doesn't expose model/providerID: gate Phase 2 on `opencode.json` being configured to use `claudecli/*` as the default model, AND on the user setting `AIDEVOPS_CLAUDE_PROXY=1`.

**(3) Adopt-on-EADDRINUSE retry (DEFENSE IN DEPTH — for two interactive sessions that both legitimately use claudecli/*).** Keep the original t3212 design as a fallback inside `ensureClaudeProxyStarted()`:

```js
async function probeExistingProxyWithRetry(maxAttempts = 5, intervalMs = 1000, timeoutMs = 1500) {
  for (let i = 0; i < maxAttempts; i++) {
    const port = await probeExistingProxy(timeoutMs);
    if (port) return port;
    if (i < maxAttempts - 1) await new Promise(r => setTimeout(r, intervalMs));
  }
  return null;
}

// In ensureClaudeProxyStarted after Bun.serve throws:
} catch (err) {
  const msg = err?.message || String(err);
  const isAddrInUse = /EADDRINUSE|address already in use|Failed to start server/i.test(msg);
  if (isAddrInUse) {
    const adoptedPort = await probeExistingProxyWithRetry();
    if (adoptedPort) {
      proxyPort = adoptedPort;
      console.error(`[aidevops] Claude proxy: adopted existing server on port ${adoptedPort} (after bind retry)`);
      return { port: adoptedPort, models: getClaudeProxyModels() };
    }
  }
  console.error(`[aidevops] Claude proxy: failed to start: ${msg}`);
}
```

Make the probe timeout configurable via `CLAUDE_PROXY_PROBE_TIMEOUT_MS` env (default 1500ms).

### Verification

```bash
# Lint
node --check .agents/plugins/opencode-aidevops/claude-proxy.mjs
node --check .agents/plugins/opencode-aidevops/index.mjs

# Headless skip — pulse workers shouldn't start the proxy
AIDEVOPS_HEADLESS=1 opencode run --print-logs --log-level WARN "test" 2>&1 | grep -c "Claude proxy" 
# Expected: 0 (no proxy startup in headless mode)

# Lazy start in interactive — proxy doesn't start until claudecli/* is picked
opencode  # start interactive, pick anthropic/claude-sonnet-4-6 model, send a message
# Expected: no "Claude proxy started on port 32125" log
opencode  # start interactive, pick claudecli/claude-opus-4-6, send a message
# Expected: "Claude proxy started on port 32125" log just before the first claudecli request

# Adopt-on-EADDRINUSE — second interactive session adopts the first one's proxy
# Run two interactive OpenCodes, both pick claudecli/* models
# Expected: second one prints "adopted existing server on port 32125" (no scary error)

# Negative test (genuine port poison)
python3 -m http.server 32125 &
sleep 1
opencode  # interactive, select claudecli/*
# Expected: original failure message after 5 retries (we don't hide real errors)
```

## Acceptance

- Headless workers (`AIDEVOPS_HEADLESS=1` or similar) NEVER start the Claude CLI proxy and NEVER print the "Failed to start server" error.
- Interactive sessions that don't select `claudecli/*` don't start the proxy either (no startup cost when the user is on Anthropic API models).
- The first `claudecli/*` request in an interactive session triggers proxy start (with at most ~5s startup latency on cold start).
- A second interactive session that legitimately uses claudecli/* adopts the first one's proxy and prints `adopted existing server on port 32125 (after bind retry)`.
- Genuine port-poison cases (non-claude-proxy listener on 32125) still produce the original error after retries — we don't hide real failures.
- Probe timeout configurable via `CLAUDE_PROXY_PROBE_TIMEOUT_MS` (default 1500ms).
- `node --check` passes on all modified files.

## Session Origin

Filed during interactive session — user reported seeing the error on every OpenCode startup. Reproduced live: port 32125 bound by a pulse worker in a managed private repo, `curl --max-time 2` timed out, 16 concurrent OpenCode instances on host (1 interactive + 15 headless pulse workers). User correctly identified that lazy-load is the architectural fix, not the EADDRINUSE retry alone.

---
[aidevops.sh](https://aidevops.sh) v3.13.16 plugin for [OpenCode](https://opencode.ai) v1.14.30 with claude-opus-4-7 spent 15m and 15,910 tokens on this with the user in an interactive session.

marcusquinn:
## Summary

Resolves #21944. Three layered, orthogonal fixes for the multi-instance OpenCode startup failure mode where N concurrent sessions print scary `EADDRINUSE` errors then fall back successfully — see issue body for the original three root causes.

## Layer 1: Skip Claude CLI proxy startup in headless workers

`index.mjs` — biggest source of waste. Pulse-spawned, GitHub Actions, and full-loop workers always target `anthropic/*` via the OAuth pool, never `claudecli/*`, so the proxy is pure dead weight there. The eager `startClaudeProxy(client, directory)` call at plugin init is removed; a new `isHeadless()` helper guards lazy-start using the canonical 4-env-var set per `AGENTS.md` t1990: `FULL_LOOP_HEADLESS`, `AIDEVOPS_HEADLESS`, `OPENCODE_HEADLESS`, `GITHUB_ACTIONS`.

## Layer 2: Lazy-start on first `claudecli/*` request

`index.mjs` — interactive sessions that never select a `claudecli/*` model also waste resources starting the proxy. Now started inside a composed `experimental.chat.system.transform` hook that delegates to TTSR after lazy-start. Pattern modelled on `ttsr.mjs` use of `input.model.providerID`. Idempotent — proxy module already caches `proxyPort` and adopts existing servers — so per-request cost is nil after the first start.

`config-hook.mjs` — `registerClaudeCliModels` no longer deletes `config.provider.claudecli` when the port is null. Without this fix, deferred startup would silently strip `claudecli/*` models from the picker. The eager `registerAnthropicModels` registration (with hardcoded `127.0.0.1:32125`) stays intact and is sufficient for the picker until the proxy actually starts.

## Layer 3: Adopt-on-EADDRINUSE retry path

`claude-proxy.mjs` — defense in depth for the residual two-interactive-sessions-at-once case where both probe (find nothing), then both call `Bun.serve` (one wins, one gets `EADDRINUSE`). New `probeExistingProxyWithRetry()` runs 5 attempts × 1s spacing; new `isPortInUseError()` heuristic fires the retry path inside `startClaudeProxy`'s catch. The sibling plugin's fetch listener may not be reachable on the first probe even though `bind()` succeeded, so the retry collapses what was previously a noisy `failed to start` log into a clean `adopted sibling server on port 32125 after EADDRINUSE` line.

Probe timeout is also now configurable: `CLAUDE_PROXY_PROBE_TIMEOUT_MS` env var, default `1500ms`. Was hardcoded `1000ms` — verified too short when an existing proxy is mid-SSE-stream (it can take >1s to answer `/v1/models`).

## Files changed

- `claude-proxy.mjs` — new constants, `probeExistingProxyWithRetry`, `isPortInUseError`, EADDRINUSE-adopt branch in `startClaudeProxy` catch.
- `index.mjs` — `isHeadless()` helper, removed eager `startClaudeProxy` call, composed `systemTransformHook` wrapper.
- `config-hook.mjs` — `registerClaudeCliModels` preserves provider entry when port is null.

## Testing

- `node --check` clean on all three modules.
- Manual flow trace confirms the three layers compose correctly: headless workers short-circuit before any proxy code runs; interactive sessions hit the `system.transform` wrapper only on `claudecli/*` requests; `Bun.serve` failures route through the retry-adopt path.
- The three fixes are mutually independent — Layer 1's behaviour change is verifiable by env-var inspection alone (no proxy code reached); Layer 2's idempotency relies on the existing `proxyPort` cache (line 380-381, unchanged); Layer 3 only changes the catch-branch of an existing failure path while the happy-path adoption via `probeExistingProxy` (line 386, unchanged) remains.

### Self-Assessed Runtime Testing

`tier:standard` plugin code, `self-assessed`. Live runtime reproduction of the original 16-session race is not feasible in a worker session. Verification limited to `node --check` + scope isolation + flow trace. The CodeRabbit / Gemini / Codacy reviewers will catch any structural issues; runtime smoke under real OpenCode load can be a follow-up issue if anything regresses.

## Trade-offs considered

- **Headless detection by env-var vs. by capability**: env-var matches the canonical set used elsewhere in the framework (see AGENTS.md t1990); a worker that wanted Claude CLI proxy would set `OPENCODE_HEADLESS=0`, which is consistent.
- **Lazy-start in `system.transform` vs `messages.transform`**: `system.transform` has `input.model.providerID` (verified at `ttsr.mjs:168`), runs before HTTP request, returns Promise so OpenCode awaits proxy readiness. `messages.transform` would also work but has weaker provider context.
- **Retry budget (5 × 1s = 5s)**: long enough for any reasonable sibling startup, short enough that genuine port poisoning still surfaces a clear error within session-startup time.

## Out of scope

- The known drift between `claude-proxy.mjs::getClaudeProxyModels()` and `model-limits.mjs::CLAUDE_MODEL_LIMITS` (former hardcodes 4-5/4-6 IDs, latter is the source of truth). Filing as a follow-up — keeps this PR's blast radius small.

---
[aidevops.sh](https://aidevops.sh) v3.13.16 plugin for [OpenCode](https://opencode.ai) v1.14.30 with claude-opus-4-7 spent 12m and 41,016 tokens on this with the user in an interactive session.

### DIFF
--- a/.agents/plugins/opencode-aidevops/claude-proxy.mjs
+++ b/.agents/plugins/opencode-aidevops/claude-proxy.mjs
@@ -46,6 +46,20 @@ const CLAUDE_PROVIDER_ID = "claudecli";
 const OPENCODE_CONFIG_PATH = join(homedir(), ".config", "opencode", "opencode.json");
 /** Opt-in debug request dump — set CLAUDE_PROXY_DEBUG_DUMP=1 to enable. */
 const DEBUG_DUMP_ENABLED = process.env.CLAUDE_PROXY_DEBUG_DUMP === "1";
+
+// Probe-existing-proxy timeout. The historical 1s default was too short when
+// the existing proxy was mid-SSE-stream (verified: a busy proxy on 32125
+// timed out a `curl --max-time 2 /v1/models` request). 1500ms is the new
+// default; override via CLAUDE_PROXY_PROBE_TIMEOUT_MS for slow hosts. See
+// GH#21944 ("multi-instance startup printing scary error") for context.
+const CLAUDE_PROXY_PROBE_TIMEOUT_MS = parseInt(
+  process.env.CLAUDE_PROXY_PROBE_TIMEOUT_MS || "1500",
+  10,
+);
+/** Retry attempts for adopt-on-EADDRINUSE — short by design; if 5 probes
+ * fail to find a server, the port is genuinely poisoned by something else. */
+const CLAUDE_PROXY_PROBE_RETRY_ATTEMPTS = 5;
+const CLAUDE_PROXY_PROBE_RETRY_INTERVAL_MS = 1000;
 const SSE_HEADERS = {
   "Content-Type": "text/event-stream",
   "Cache-Control": "no-cache",
@@ -306,16 +320,24 @@ async function registerProxyAuth(client) {
 
 /**
  * Probe whether a proxy server is already listening on CLAUDE_PROXY_DEFAULT_PORT.
- * Used to adopt an existing server after plugin hot-reload (where the module
- * scope resets but the Bun.serve instance from the previous load lives on) or
- * when the plugin runs in a non-Bun JS runtime that cannot call Bun.serve.
+ * Used in three scenarios:
+ *   1. Pre-launch: avoid double-binding when another OpenCode session already
+ *      runs a proxy on the default port.
+ *   2. After plugin hot-reload: the module scope resets but the Bun.serve
+ *      instance from the previous load lives on.
+ *   3. Adopt-on-EADDRINUSE: when our own `Bun.serve` raises EADDRINUSE, the
+ *      port owner is virtually always a sibling proxy that's still mid-startup
+ *      (its fetch listener may not be ready yet — see probeExistingProxyWithRetry).
+ *
+ * Timeout sourced from CLAUDE_PROXY_PROBE_TIMEOUT_MS (default 1500ms) — a busy
+ * proxy mid-SSE-stream can take >1s to answer a /v1/models request.
  *
  * Returns the port number on success, null if no server is reachable.
  */
 async function probeExistingProxy() {
   try {
     const controller = new AbortController();
-    const timer = setTimeout(() => controller.abort(), 1000);
+    const timer = setTimeout(() => controller.abort(), CLAUDE_PROXY_PROBE_TIMEOUT_MS);
     const res = await fetch(
       `http://127.0.0.1:${CLAUDE_PROXY_DEFAULT_PORT}/v1/models`,
       { signal: controller.signal },
@@ -326,6 +348,42 @@ async function probeExistingProxy() {
     // not running or not reachable
   }
   return null;
+}
+
+/**
+ * Probe with bounded retries — used by the EADDRINUSE adoption path. When a
+ * sibling plugin is mid-startup, the port may already be `bind()`ed by Bun
+ * but the fetch listener not yet reachable, so a single probe returns null
+ * even though the owner will be ready in <1s.
+ *
+ * Retries CLAUDE_PROXY_PROBE_RETRY_ATTEMPTS times with
+ * CLAUDE_PROXY_PROBE_RETRY_INTERVAL_MS spacing. Returns the port on first
+ * success, null if every attempt fails.
+ */
+async function probeExistingProxyWithRetry() {
+  for (let attempt = 0; attempt < CLAUDE_PROXY_PROBE_RETRY_ATTEMPTS; attempt++) {
+    const port = await probeExistingProxy();
+    if (port) return port;
+    if (attempt < CLAUDE_PROXY_PROBE_RETRY_ATTEMPTS - 1) {
+      await new Promise((resolve) =>
+        setTimeout(resolve, CLAUDE_PROXY_PROBE_RETRY_INTERVAL_MS),
+      );
+    }
+  }
+  return null;
+}
+
+/**
+ * Heuristic: is this error a port-already-bound failure rather than a deeper
+ * Bun.serve config problem? Bun raises EADDRINUSE; some shims raise
+ * `listen EADDRINUSE`. We check both `code` and `message` so we don't miss
+ * either form.
+ */
+function isPortInUseError(err) {
+  if (!err) return false;
+  if (err.code === "EADDRINUSE") return true;
+  const msg = typeof err.message === "string" ? err.message : "";
+  return /EADDRINUSE|address already in use/i.test(msg);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


alex-solovyev (approved):
Auto-approved by pulse runner @alex-solovyev — author @marcusquinn confirmed collaborator, pre-merge gates passed.

PR #23287, opened 2026-05-09
fix: avoid read-before-edit TTSR false positives (GH#23226)

### LINKED ISSUE #23226: bug: TTSR read-before-edit rule still fires false positives — #2487 lookahead too narrow; rule is structurally redundant with native Edit/Write enforcement
## Description

The `read-before-edit` TTSR rule (`.agents/plugins/opencode-aidevops/ttsr-rules.mjs:31-38`) still fires false-positive `<!-- TTSR:ERROR:read-before-edit -->` markers in the assistant's `text.complete` output.

Issue #2487 was closed `status:done` after adding the negative lookahead `(?!.*(?:creat|new file|new \w+ file|generat))` to suppress the rule on new-file creation. The lookahead is too narrow: it only matches four stems (`creat`, `new file`, `new \w+ file`, `generat`) anywhere in the rest of the string. In practice, almost no natural assistant phrasing contains those exact stems, so the rule fires on:

- prose that has nothing to do with file ops, e.g. discussion of the rule itself ("Let me write the brief", "Let me write up what I found")
- legitimate Edits that *were* preceded by Read tool calls, because the assistant narrated the edit without using the literal phrases "I'll read", "let me read", "I've read", or "already read"
- new-file Writes whose narration uses words other than the four stems, e.g. "Let me write a unit test", "Let me write a body file", "Let me write the implementation"

The rule cannot reach correctness as a pure text regex — it has no access to tool history, so it cannot distinguish (a) prose containing the trigger phrase, (b) Edits with a prior Read tool call, (c) Writes to new files, and (d) Writes to existing files without a prior Read. Only case (d) is a true violation, and the Edit/Write tools themselves already enforce that with a hard error ("File has not been read yet").

## Expected Behavior

Either:

1. The rule's matcher consults tool history (`session.tools` or equivalent) to suppress the marker when (a) no Edit/Write occurred in the same turn, or (b) the target file was Read earlier in the conversation, or (c) the target is a new file — making the rule actually correct; or
2. The rule is removed, since the Edit/Write tools self-enforce read-before-edit at the tool boundary with higher fidelity than a regex can achieve.

A weaker interim option is to drop severity from `error` to `warn` and broaden the lookahead window so common natural-language phrasing stops triggering it — but option 2 is preferable because the rule is structurally redundant with native tool behaviour.

## Reproducer

**Symptom command** (this very turn — the assistant performed only Read and Bash tool calls, no Edit, no Write):

````
<assistant text containing the phrase "Let me write the brief">
````

**Actual output** (injected by `text.complete` hook at `ttsr.mjs:325`):

```
<!-- TTSR:ERROR:read-before-edit — ALWAYS Read a file before Edit/Write to an existing file. These tools fail without a prior Read in this conversation. (This rule does not apply when creating new files.) -->
```

**Expected output**: no marker. No Edit/Write was attempted; the trigger phrase appeared in unrelated narrative prose.

**Causal code**:

```
de601e9347 (alex-solovyev 2026-04-14 03:31:32 -0400 34)     pattern: "(?:I'll edit|Let me edit|I'll write to|Let me write)(?!.*(?:creat|new file|new \\w+ file|generat))(?:(?!I'll read|let me read|I've read|already read).){0,200}$",
```

The lookahead `(?!.*(?:creat|new file|new \w+ file|generat))` was introduced by the fix for #2487 but does not cover the false-positive surface described there or here.

**Call-site sweep**:

```
.agents/plugins/opencode-aidevops/ttsr-rules.mjs:31-38   # rule definition
.agents/plugins/opencode-aidevops/ttsr.mjs:303-325        # text.complete hook that emits the marker
```

## Steps to Reproduce

1. In an OpenCode session with the `opencode-aidevops` plugin loaded, ask the assistant to summarise something using prose like "Let me write up what I found", with no Edit/Write tool call.
2. Inspect the rendered assistant message: an `<!-- TTSR:ERROR:read-before-edit ... -->` marker is appended even though no edit was attempted.
3. Alternatively, perform an Edit that was preceded by a Read tool call but narrate it with "Now let me edit the file": the marker still appears, because the regex looks for literal "I'll read"/"I've read" phrasing in the same text window, not in tool history.

## Workarounds Applied

During this session:

- The assistant rephrased prose to avoid the trigger phrases (avoiding "Let me write" / "Let me edit"), which is prompt theatre that doesn't change underlying behaviour.
- The user manually ignored the markers, knowing they were spurious. This pollutes the conversation with noisy `<!-- TTSR:ERROR -->` HTML comments and erodes signal value when a real violation eventually appears.

There is no clean per-session escape — the rule is built-in (`ttsr-rules.mjs`) and not user-overridable without editing the deployed plugin file.

## Environment

- aidevops version: 3.15.9
- Latest version: 3.15.9
- Install method: unknown (/home/dave/.local/bin/aidevops)
- AI Assistant: OpenCode
- OS: Ubuntu 24.04.4 LTS
- Shell: bash 5.2.21(1)-release
- Working repo: multisite-ultimate-domain-seller (main)
- gh CLI: gh version 2.45.0

## Additional Context

- Prior fix attempt: #2487 (closed `status:done` 2026-02-27). That fix updated `prompts/build.txt` wording and added the negative lookahead to the regex. The wording fix landed correctly; the regex fix is structurally insufficient.
- Related decomposition: `de601e9347` (t2071, #18906) extracted rules into `ttsr-rules.mjs` without altering the regex.
- Native enforcement: the Edit and Write tool prompts already require a prior Read for existing files and produce a hard error ("File has not been read yet") if violated. The TTSR rule duplicates this with strictly lower fidelity.
- Recommended direction: convert `read-before-edit` into a tool-history-aware check inside `ttsr.mjs` (looking up the most recent Edit/Write tool call and its target file, then checking whether a Read of that path occurred earlier in the session), or remove the rule entirely. A pure-regex implementation cannot achieve correctness here.

---
[aidevops.sh](https://aidevops.sh) v3.15.9 plugin for [OpenCode](https://opencode.ai) v1.14.41 with claude-opus-4-7 spent 1h 3m and 62,678 tokens on this with the user in an interactive session.

marcusquinn:
## Summary
- Neutralizes the text-only `read-before-edit` TTSR regex so plain assistant prose no longer creates false-positive error markers.
- Keeps the system-prompt guidance intact so the invariant remains visible while enforcement stays with tool-aware hooks.
- Adds regression coverage for GH#23226.

## Verification
- `node --test .agents/plugins/opencode-aidevops/tests/test-ttsr-read-before-edit.mjs` — pass
- `git diff --check` — pass
- `node --test .agents/plugins/opencode-aidevops/tests/*.mjs` — unrelated existing failures in `test-memory-tool-schema.mjs` (`z.number is not a function`); new TTSR test passed in that run
- `.agents/scripts/linters-local.sh` — timed out after existing advisory findings; no changed-file shell or secret regressions reported before timeout

Resolves #23226

---
[aidevops.sh](https://aidevops.sh) v3.15.15 plugin for [OpenCode](https://opencode.ai) v1.14.44 with gpt-5.5 spent 3h 37m and 93,629 tokens on this with the user in an interactive session. Solved in 1d 3h.

### DIFF
--- a/.agents/plugins/opencode-aidevops/tests/test-ttsr-read-before-edit.mjs
+++ b/.agents/plugins/opencode-aidevops/tests/test-ttsr-read-before-edit.mjs
@@ -1,0 +1,41 @@
+// SPDX-License-Identifier: MIT
+// SPDX-FileCopyrightText: 2025-2026 Marcus Quinn
+//
+// Regression coverage for GH#23226: read-before-edit is a tool-history
+// invariant, not a reliable text-output regex.
+//
+//   node --test .agents/plugins/opencode-aidevops/tests/test-ttsr-read-before-edit.mjs
+
+import { test, describe } from "node:test";
+import assert from "node:assert/strict";
+
+import { BUILTIN_TTSR_RULES, scanForViolations } from "../ttsr-rules.mjs";
+
+function createState() {
+  return {
+    ttsrRules: [...BUILTIN_TTSR_RULES],
+    readIfExists: () => "",
+  };
+}
+
+describe("TTSR read-before-edit rule", () => {
+  test("keeps the read-before-edit instruction in system rules", () => {
+    const rule = BUILTIN_TTSR_RULES.find((candidate) => candidate.id === "read-before-edit");
+
+    assert.ok(rule, "read-before-edit rule must remain available");
+    assert.match(rule.systemPrompt, /ALWAYS Read a file before Edit or Write/);
+  });
+
+  test("does not flag edit intent from plain output text", () => {
+    const violations = scanForViolations(
+      "I'll edit the existing file now that I have read it.",
+      createState(),
+    );
+
+    assert.deepEqual(
+      violations.filter((violation) => violation.rule.id === "read-before-edit"),
+      [],
+      "text-only read-before-edit regex must not fire without tool-history context",
+    );
+  });
+});
--- a/.agents/plugins/opencode-aidevops/ttsr-rules.mjs
+++ b/.agents/plugins/opencode-aidevops/ttsr-rules.mjs
@@ -31,7 +31,9 @@ export const BUILTIN_TTSR_RULES = [
   {
     id: "read-before-edit",
     description: "Always Read a file before Edit or Write to existing files",
-    pattern: "(?:I'll edit|Let me edit|I'll write to|Let me write)(?!.*(?:creat|new file|new \\w+ file|generat))(?:(?!I'll read|let me read|I've read|already read).){0,200}$",
+    // Text-only detection cannot know whether a Read tool call already happened,
+    // so keep this as a system-prompt rule and leave enforcement to tool hooks.
+    pattern: "(?!)",
     correction: "ALWAYS Read a file before Edit/Write to an existing file. These tools fail without a prior Read in this conversation. (This rule does not apply when creating new files.)",
     severity: "error",
     systemPrompt: "ALWAYS Read a file before Edit or Write to an existing file. These tools FAIL without a prior Read in this conversation. For NEW files, verify the parent directory exists instead.",


alex-solovyev:
Permission check failed for this PR (HTTP unknown from collaborator permission API). Unable to determine if @marcusquinn is a maintainer or external contributor. **A maintainer must review and merge this PR manually.** This is a fail-closed safety measure — the pulse will not auto-merge until the permission API succeeds.

---
[aidevops.sh](https://aidevops.sh) v3.15.16 plugin for [OpenCode](https://opencode.ai) v1.14.41 with gpt-5.5 spent 4m and 36,476 tokens on this as a headless worker.

PR #30594, opened 2026-08-23
GH#30575: restore local model upstream compatibility

### LINKED ISSUE #30575: local-model-helper.sh: 6 version-drift breaks vs current llama.cpp / huggingface_hub releases
## Summary

`local-model-helper.sh` / `local-model-setup.sh` / `local-model-server.sh` /
`local-model-models.sh` (the llama.cpp-based local inference stack) have six
version-drift compatibility breaks against current upstream releases of
`ggml-org/llama.cpp` and `huggingface_hub`. Hit all six in one session
running `local-model-helper.sh setup` → `download` → `start` on a clean
Linux x64 CPU-only host (no prior llama.cpp/hf installs). Each is
reproducible today; details, evidence, and suggested fixes below.

Environment: aidevops v3.32.287→v3.32.288 (auto-updated mid-session),
Linux x64, no GPU, bash.

---

### 1. `local-model-setup.sh` — release asset resolution fails (llama.cpp semver tags)

`LLAMA_CPP_API` hits `https://api.github.com/repos/ggml-org/llama.cpp/releases/latest`.
llama.cpp recently introduced stable semver tags (`vX.Y.Z`) that are
now what `/releases/latest` returns — but that release only ships a
`nightly-tag.txt` pointer asset, not prebuilt binaries. The actual binaries
live on a separate nightly `b<NUM>` tag referenced inside that file.

Repro:
```
$ local-model-helper.sh setup
...
[INFO] Latest release: v0.2.0
[ERROR] No matching release asset for pattern: llama-.*-bin-ubuntu-x64\.tar\.gz
```

```
$ curl -s https://api.github.com/repos/ggml-org/llama.cpp/releases/latest | jq '.assets[].name'
"nightly-tag.txt"
```

Fix direction: when the resolved release's assets don't match the expected
pattern, fetch the `nightly-tag.txt` asset content (a plain-text tag name,
e.g. `b10566`) and re-resolve against `releases/tags/<that-tag>` instead.

### 2. `local-model-setup.sh:_setup_install_binaries` — only copies 2 binaries, drops required shared libraries

Modern llama.cpp release archives are dynamically linked — `llama-server`
depends on ~15 co-located `.so` files (`libggml-base.so.0`, `libllama.so.0`,
`libmtmd.so.0`, etc.), confirmed via `ldd`. `_setup_install_binaries` only
copies `llama-server` and `llama-cli` into `LOCAL_BIN_DIR`, leaving the
`.so` dependencies behind in the extracted temp dir — the copied binary
would fail to run once the temp dir is cleaned up (untested directly since
issue #1 blocked reaching this step, but confirmed via `ldd` output showing
all deps resolve to sibling files in the same directory as the binary,
consistent with `$ORIGIN`-relative rpath).

Fix direction: copy the entire extracted release directory's contents into
`LOCAL_BIN_DIR`, not just the two named binaries.

### 3. `local-model-helper.sh status` — doesn't parse current `--version` output

```
$ /path/to/llama-server --version
version: 0.2.0-dev (build 10566, commit bb4caa754)
built with GNU 11.4.0 for Linux x86_64
```
```
$ local-model-helper.sh status
Server: not running
Binary:
```
"Binary:" renders empty even with a working, executable binary in place.
Likely a regex/parsing assumption tied to the older `b<NUM>`-only version
string format.

### 4. `huggingface-cli` binary renamed to `hf`

Current `huggingface_hub` (tested: 1.28.0) ships only the `hf` command.
`local-model-setup.sh`'s installer and `local-model-models.sh`'s dependency
check both reference `huggingface-cli`, which no longer exists after a
fresh `pip install "huggingface_hub[cli]"`. Also note: that pip invocation
itself now warns `huggingface-hub 1.28.0 does not provide the extra 'cli'`
(the CLI is bundled by default; the extra is a no-op on current versions).

Fix direction: check for `hf` first, fall back to `huggingface-cli` for
older installs; drop the `[cli]` extra from the pip install command.

### 5. `local-model-models.sh:cmd_download` — `--local-dir-use-symlinks` flag removed

```
huggingface-cli download "$repo" "$filename" \
    --local-dir "$LOCAL_MODELS_STORE" \
    --local-dir-use-symlinks False
```
`hf download --help` (current) has no `--local-dir-use-symlinks` flag —
removed in the `huggingface_hub` v1.x CLI rewrite. This flag would cause an
immediate argument-parse failure once issue #4 is fixed and the real `hf`
binary is invoked.

### 6. `local-model-server.sh:_start_server_process` — bare `--flash-attn` flag rejected

```
server_args+=("--flash-attn")   # no value
```
Current `llama-server` requires an explicit value:
```
error while handling argument "--flash-attn": expected value for argument
usage: -fa, --flash-attn [on|off|auto]  (default: 'auto')
```
`--no-flash-attn` (a separately supported CLI override in
`local-model-helper.sh`) works around this by skipping the flag entirely,
but the default `flash_attn="true"` path is broken for every user who
doesn't pass that override.

Fix direction: `server_args+=("--flash-attn" "auto")` (or "on") instead of
the bare flag.

---

## Verification

Worked around all six manually (fetch nightly-tag.txt → resolve real
release, copy full extracted dir, shim `hf`→`huggingface-cli`, download via
`hf` directly without the removed flag, start with `--no-flash-attn`) and
confirmed a working end-to-end local inference server: `Qwen3-14B-Q4_K_M`
loaded via `llama-server` on `127.0.0.1:8080`, verified with a real
`/v1/chat/completions` call.

---
[aidevops.sh](https://aidevops.sh) v3.32.288 plugin for [OpenCode](https://opencode.ai) v1.18.21 with claude-sonnet-5

marcusquinn:
## Summary

- resolve llama.cpp stable pointer releases to their binary-bearing nightly tag
- atomically install the complete release directory, retaining required shared libraries and restoring the prior directory on validation failure
- support the current `hf` CLI in an isolated environment while retaining legacy CLI compatibility
- normalize current version output, download flags, Flash Attention syntax, and automatic GPU-layer defaults
- update local-model documentation for the current runtime and CLI layout

## Verification

- `.agents/scripts/linters-local.sh`
- focused release-resolution, version-matching, complete-directory install, and rollback fixtures
- clean temporary-home setup installed nightly `b10566` and the isolated `hf` CLI
- downloaded `TinyMistral-248M-v2-Instruct.Q4_K_M.gguf`, started `llama-server`, completed a real `/v1/chat/completions` request, and stopped cleanly
- `setup --update` replaced the complete llama.cpp directory while preserving the working isolated `hf` environment

Resolves #30575

---
[aidevops.sh](https://aidevops.sh) v3.32.288 plugin for [OpenCode](https://opencode.ai) v1.18.19 with gpt-5.6-sol spent 1h 1m and 558,863 tokens on this with the user in an interactive session.

### DIFF
--- a/.agents/scripts/local-model-db.sh
+++ b/.agents/scripts/local-model-db.sh
@@ -89,7 +89,7 @@ load_config() {
 		LLAMA_PORT="$(jq -r '.port // 8080' "$LOCAL_CONFIG_FILE" 2>/dev/null || echo "8080")"
 		LLAMA_HOST="$(jq -r '.host // "127.0.0.1"' "$LOCAL_CONFIG_FILE" 2>/dev/null || echo "127.0.0.1")"
 		LLAMA_CTX_SIZE="$(jq -r '.ctx_size // 8192' "$LOCAL_CONFIG_FILE" 2>/dev/null || echo "8192")"
-		LLAMA_GPU_LAYERS="$(jq -r '.gpu_layers // 99' "$LOCAL_CONFIG_FILE" 2>/dev/null || echo "99")"
+		LLAMA_GPU_LAYERS="$(jq -r '.gpu_layers // "auto"' "$LOCAL_CONFIG_FILE" 2>/dev/null || echo "auto")"
 		LLAMA_FLASH_ATTN="$(jq -r '.flash_attn // true' "$LOCAL_CONFIG_FILE" 2>/dev/null || echo "true")"
 	fi
 	return 0
@@ -104,7 +104,7 @@ write_default_config() {
 			  "host": "127.0.0.1",
 			  "ctx_size": 8192,
 			  "threads": "auto",
-			  "gpu_layers": 99,
+			  "gpu_layers": "auto",
 			  "flash_attn": true
 			}
 		CONFIGEOF
--- a/.agents/scripts/local-model-helper.sh
+++ b/.agents/scripts/local-model-helper.sh
@@ -9,7 +9,7 @@
 # Usage: local-model-helper.sh [command] [options]
 #
 # Commands:
-#   install [--update]            Install/update llama.cpp + huggingface-cli (alias: setup)
+#   install [--update]            Install/update llama.cpp + Hugging Face CLI (alias: setup)
 #   serve [--model M] [options]   Start llama-server localhost:8080 (alias: start)
 #   stop                          Stop running llama-server
 #   status                        Show server status and loaded model
@@ -29,7 +29,7 @@
 #   --port N        Server port (default: 8080)
 #   --ctx-size N    Context window size (default: 8192)
 #   --threads N     CPU threads (default: auto-detect performance cores)
-#   --gpu-layers N  GPU layers to offload (default: 99 = all)
+#   --gpu-layers N  GPU layers to offload (default: auto)
 #   --json          Output in JSON format
 #   --quiet         Suppress informational output
 #
@@ -44,7 +44,7 @@
 # Exit codes:
 #   0 - Success
 #   1 - General error
-#   2 - Dependency missing (llama.cpp, huggingface-cli)
+#   2 - Dependency missing (llama.cpp, Hugging Face CLI)
 #   3 - Model not found
 #   4 - Server already running / not running
 #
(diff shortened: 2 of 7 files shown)


marcusquinn:
## Merge summary

- Restored compatibility with current llama.cpp release pointers, shared-library bundles, version output, and server flags.
- Added current `hf` support with an isolated installation and retained legacy CLI compatibility.
- Verified clean setup, model download, server start, a real chat-completions request, update behavior, and failed-install rollback.
- All required CI and review gates passed before squash merge.

Merged as `e01c4d36163ede0c2a1c4ca97eabacac18e3c610`.

Release: not requested.



---
[aidevops.sh](https://aidevops.sh) v3.32.288 plugin for [OpenCode](https://opencode.ai) v1.18.19 with gpt-5.6-sol spent 1h 20m and 616,382 tokens on this with the user in an interactive session.

PR #31046, opened 2026-09-02
GH#31045: fix: trust fast-uri Dependabot update

vladimirdulov:
## Summary

Allowlist the reviewed npm_and_yarn fast-uri update and apply Dependabot's exact 3.1.7 lockfile integrity data.

## Files Changed

.agents/configs/trusted-dependabot-updates.conf, .agents/package-lock.json

## Runtime Testing

- **Risk level:** Medium
- **Verification:** self-assessed — bash .agents/scripts/tests/test-pulse-merge-trusted-dependabot.sh; npm ci --prefix .agents --ignore-scripts; npm ls --prefix .agents fast-uri; npm audit --prefix .agents --omit=dev --audit-level=high

Resolves #31045

---
[aidevops.sh](https://aidevops.sh) v3.32.298 plugin for [OpenCode](https://opencode.ai) v1.18.18 with gpt-5.6-terra spent 3m and 104,827 tokens on this as a headless worker.

Issue #31027, opened 2026-09-02
fix(storage): retrofit safe cache pruning for protected mixed recovery archives


katarzynacc:
## Problem

PR #31021 correctly recognises safe Python and repository-root CodeGraph caches for new archive pruning and cache-only classification, but existing mixed archives retain those cache bytes forever whenever any protected ignored artifact remains. The terminal archive stays dirty, so maintenance cannot delete the bucket, and there is no sanctioned retrofit operation that removes only the newly approved cache roots while preserving protected content and archive evidence.

## Runtime evidence after #31021

Deployed commit 82c1aaad3 was exercised through one complete bounded maintenance cycle on the affected Linux host. The run scanned 332/332 buckets with deadline_exhausted=false and sizing failures=0, but selected zero candidates and reclaimed zero bytes. Reasons: 124 archive-worktree-dirty, 181 active-session-claim, 19 required-evidence-unavailable, and 8 unknown archives. Filesystem remained at 90 percent usage with 13 GiB free.

A read-only allocation scan found 14,052,290,560 bytes under cache roots already approved by #31021 inside the recovery store: 13,618,311,168 bytes across 149 .codegraph roots, 427,307,008 bytes across 2,486 __pycache__ roots, and 6,672,384 bytes across 65 .pytest_cache roots. A representative mixed archive had a 103 MiB root .codegraph plus tiny protected logs; current classification correctly preserves the logs but consequently retains the entire cache too.

This is not a request to broaden disposable-path classification. Arbitrary logs, databases outside explicit cache roots, tracked files, nested .codegraph, symlinks, and user artifacts must remain protected.

## Files

- .agents/scripts/worktree-recovery-cache-policy.py
- .agents/scripts/worktree-recovery-maintenance-helper.sh
- .agents/scripts/worktree-recovery-apply-helper.sh if transactional mutation authority belongs there
- .agents/scripts/tests/test-worktree-recovery-lifecycle.sh
- .agents/reference/storage-lifecycle.md

## Required change

Add a bounded, journaled, fail-closed retrofit maintenance path that prunes only cache roots already proven safe by worktree-recovery-cache-policy.py from existing recovery archive copies, even when other protected data keeps the bucket non-candidate. Revalidate archive identity, ordinary-directory and symlink guards, tracked-file absence, and exact allocated bytes immediately before mutation. Preserve remaining archive content and update or regenerate any integrity evidence required for future recovery classification rather than leaving stale metadata.

[...]

PR #31051, opened 2026-09-03
GH#31027: aggregate v3.32.300 release sources

marcusquinn:
## Summary

- prepare the reviewed exact-tip aggregation required for v3.32.300
- bind unreleased merged sources #31046 and #31050 to immutable merge commits
- preserve the current source tree through a metadata-only squash merge

Ref #31027

## Expected sources

- `31046@1cf7fd24247fb6b36527a8019203f332d296179c`
- `31050@c57d45eec2dc724f31b5ecc639b19352da4984c7`

## Verification

- v3.32.299 is the latest published release
- both source PRs are merged to main with terminal-success required checks
- the initial publication attempt stopped before version, tag, or package mutation
- the final follow-up commit will carry one contiguous immutable trailer block after this PR number is allocated

## Runtime Testing

Self-assessed: metadata-only aggregation with no tree diff; repository gates validate the exact head and squash trailers.


marcusquinn:
Superseded before review: the follow-up commit separated release trailers into multiple paragraphs, so `git interpret-trailers --parse` exposed only a partial manifest. No tree changes or release side effects occurred. A fresh aggregation PR will preserve one contiguous trailer block without rewriting this published branch.

